PR #119, opened 2025-11-10
Update cmake fetch dependencies

mohsensadr:
### Changes in CMake:

In this PR, we enable opalx's cmake to build opalx's dependencies, including 

```
IPPL
HDF5
H5hut
Boost
GSL
GoogleTest
```

We restructure the cmake of OPALX similar to the way it's done for IPPL. The challenging dependencies are fetched in 

https://github.com/OPALX-project/OPALX/blob/0d13c6fa31383fe866d5074cf02a1fc5168b2ea2/cmake/Dependencies.cmake#L2

The cmake command is as before

```
mkdir build_serial && cd build_serial
cmake .. \
    -DCMAKE_BUILD_TYPE=Debug \
    -DCMAKE_CXX_STANDARD=20 \
    -DIPPL_ENABLE_FFT=ON \
    -DIPPL_ENABLE_SOLVERS=ON \
    -DIPPL_ENABLE_ALPINE=OFF \
    -DIPPL_ENABLE_TESTS=OFF  \
    -DIPPL_PLATFORMS=serial
```

and then build as before 

`make`.

I managed to build and run OPALX on my local machine (Mac with clang 17 from xcode).

### Some Changes in the Source File

- Rename src/Attributes/String.h to OpalString.h -> clang confuses String.h with string.h from C++.
- Some implementations in src/AbsBeamline/Monitor.cpp are giving error (needs to be fixed separately). The relevant code is commented out and it would throw error if they are ever called. 

I made an issue to address the errors in  src/AbsBeamline/Monitor.cpp separately  https://github.com/OPALX-project/OPALX/issues/120

### Update
It also compiles with gcc15 on Mac and Juelich Booster as reported by @rammann.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,13 +30,8 @@ include(ProjectSetup)
 # ------------------------------------------------------------------------------
 # Primary OPALX options
 # ------------------------------------------------------------------------------
-set(OPALX_PLATFORMS "SERIAL" CACHE STRING "Platforms to build OPALX for")
 option(BUILD_SHARED_LIBS "Build OPALX as a shared library" OFF)
 option(OPALX_ENABLE_UNIT_TESTS "Enable unit tests using GoogleTest" OFF)
-option(OPALX_ENABLE_FFT "Enable FFT support" OFF)
-option(OPALX_ENABLE_SOLVERS "Enable OPALX solvers" OFF)
-option(OPALX_ENABLE_ALPINE "Enable building the Alpine module" OFF)
-option(OPALX_ENABLE_COSMOLOGY "Enable building the Cosmology module" OFF)
 option(OPALX_ENABLE_EXAMPLES "Enable building the Example module" OFF)
 option(OPALX_ENABLE_TESTS "Build integration tests in test/ directory" OFF)
 option(OPALX_ENABLE_COVERAGE "Enable code coverage" OFF)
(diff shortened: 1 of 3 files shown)


mohsensadr:
On Mac, use Xcode's clang instead of gcc-15. Make sure your clang++ is the one from Xcode.

```
$ /usr/bin/clang++ --version
Apple clang version 17.0.0 (clang-1700.4.4.1)
Target: arm64-apple-darwin24.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

I had to update my Xcode, looking at the history of commands, I did things like

```
sudo xcode-select --reset
sudo xcodebuild -runFirstLaunch
sudo xcodebuild -license accept
xcrun --show-sdk-path
sudo xcode-select -s /Applications/Xcode.app/Contents/Developer
export SDKROOT=$(xcrun --sdk macosx --show-sdk-path)
brew install make
```

mohsensadr:
Yes, I can imagine using more than one thread could cause a problem with the sequence of dependencies...

mohsensadr:
Since gsl is not fetched like others, we need to explicitly say that opalx is dependent on it 

```
# make sure gsl is built before opalx
add_dependencies(opalx gsl_external)
```

So that the compiler won't try compiling opalx before gsl. 

@rammann can you try again?

Issue #120, opened 2025-11-10
Error in implementation of AbsBeamline/Monitor.cpp


mohsensadr:
In the process of rewriting the cmake, we came across these errors in AbsBeamline/Monitor.cpp that needs to be fixed. For now, we have commented them out and throw error when the functions are called:

```
/Users/mohsensadr/Codes/GitHub/OPALX/src/AbsBeamline/Monitor.cpp:68:20: error: no viable conversion from 'Vector_t<double, 3U>' (aka 'Vector<double, 3U>') to 'const double'
   68 |      const double& dt                     = RefPartBunch_m->dt(i);
      |                    ^                        ~~~~~~~~~~~~~~~~~~~~~
/Users/mohsensadr/Codes/GitHub/OPALX/src/AbsBeamline/Monitor.cpp:76:34: error: no member named 'ID' in 'PartBunch<double, 3>'
   76 |                  RefPartBunch_m->ID[i], R + frac * singleStep, P, t + frac * dt,
      |                  ~~~~~~~~~~~~~~  ^
/Users/mohsensadr/Codes/GitHub/OPALX/src/AbsBeamline/Monitor.cpp:77:34: error: no member named 'Q' in 'PartBunch<double, 3>'
   77 |                  RefPartBunch_m->Q[i], RefPartBunch_m->M[i]));
      |                  ~~~~~~~~~~~~~~  ^
/Users/mohsensadr/Codes/GitHub/OPALX/src/AbsBeamline/Monitor.cpp:77:56: error: no member named 'M' in 'PartBunch<double, 3>'
   77 |                  RefPartBunch_m->Q[i], RefPartBunch_m->M[i]));
      |                                        ~~~~~~~~~~~~~~  ^
/Users/mohsensadr/Codes/GitHub/OPALX/src/AbsBeamline/Monitor.cpp:102:33: error: invalid range expression of type 'PartBunch_t' (aka 'PartBunch<double, 3>'); no viable 'begin' function available
  102 |      for (OpalParticle particle : *RefPartBunch_m) {
      |                                 ^ ~
/Users/mohsensadr/Codes/GitHub/OPALX/src/AbsBeamline/Monitor.cpp:125:39: error: no matching function for call to 'euclidean_norm'
  125 |              double ds              = euclidean_norm(dR + 0.5 * singleStep);
      |                                       ^~~~~~~~~~~~~~
[...]

PR #127, opened 2025-11-12
Fix monitor

mohsensadr:
Rewrite apply function of monitor class given how particle attributes are currently accessed. Also, I removed declaration of apply signature that was not implemented even in the old OPAL and never used.

Here, we bring back the commented out code during rewrite of the new cmake:
https://github.com/OPALX-project/OPALX/issues/120

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -1,0 +1,92 @@
+name: CI workflow
+          
+on:
+  ## Trigger CI workflow on push to any branch (commented out to limit runs)
+  #push:
+  ## Trigger CI workflow on pull requests to the master branch   
+  pull_request:
+    branches: [ "master" ]
+          
+jobs:     
+  # ===== CPU SERIAL BUILD =====
+  build-cpu:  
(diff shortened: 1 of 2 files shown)


rammann:
As far as I can tell `RefPartBunch` is an object of type `PartBunch`. If we look at `PartBunch.h` lines 276-282 we see that the functions `PartBunch::R `and `PartBunch::P` return 0 vectors. Is this intended or simply forgotten part? If you want the Views with R and P you need `PartBunch::getParticleContainer::R` and `PartBunch::getParticleContainer::P`.

mohsensadr:
As mentioned before, In monitor we are collecting particle information using host-only implementations of the old OPAL with calls like `addParticle `from `LossDataSink `class which does dynamic allocation (`push back` of `std::vector<size_t>` for example  for `bunchNumber_m ` and `turnNumber_m ` which are not allowed on GPU.

So, I think `LossDataSink `class should be first kokkosified. Only then we can implement monitor.

I close this PR for now.

PR #128, opened 2025-11-14
Compatibility for cmake version 3.27 and earlier

klappi-s:
CMP0155 is only availlable from  cmake 3.28 and on.
Inserting the conditional policy set:

```
if(POLICY CMP0155)
cmake_policy(SET CMP0155 OLD)
endif()
```

avoids the unknown policy error, and enables successfull compilations and runs for old cmake versions.


mohsensadr:
gsl link is broken. I made another pull request to add other mirrors for it
https://github.com/OPALX-project/OPALX/pull/135

PR #133, opened 2025-11-19
CMake options to use pre-built dependencies

mohsensadr:
In this PR, we let opalx's users to use their system pre-built dependencies

- HDF5 1.14.6
- H5hut 2.0.0rc6
- Gtest 1.16.0

by setting the corresponding option to `ON` in the cmake command:

- `-DOPALX_USE_INSTALLED_HDF5=ON`
- `-DOPALX_USE_INSTALLED_H5HUT=ON`
- `-DOPALX_USE_INSTALLED_GTEST=ON`

These options are set to `OFF` by default, meaning opalx will fetch and build these dependencies by itself.

### DIFF
--- a/cmake/Dependencies.cmake
+++ b/cmake/Dependencies.cmake
@@ -260,51 +260,130 @@ if(OPALX_ENABLE_UNIT_TESTS)
         if(NOT TARGET GTest::gtest_main)
             message(STATUS "⚙ Creating GTest targets manually from variables...")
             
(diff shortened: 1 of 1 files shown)


mohsensadr:
@rammann could you try again? Now it's working on merlin 7 as well as 6 for me at least.
On merlin7, you may want to have:

`module load zstd/1.5.6-zt22`

PR #163, opened 2026-01-14
Fix Coordinate System Transformation and GPU-Compatible Beam Line Elements

rammann:
This PR consists of two main parts

## GPU Compatible Coordinate System Transformation
Before, we relied on a boost implementation for the required 3x3 matrix-vector multiplications (`src/Algorithms/BoostMatrix.h`) for coordinate system transformations. Since these transformations need to be applied to the particle bunch on device, this is not a viable option. 

`src/Algorithms/OpalMatrix.h` implements a 3x3 matrix class that can be copied to device, with functions that can be called from within a GPU kernel. 

Classes , like `Quaternion`, which previously made use of `BoostMatrix` had to be adjusted to use `OpalMatrix` instead.
Their functionality remains the same. 

In the class `CoordinateSystemTrafo`, the functions for transforming a single particle (point) are:

- `ippl::Vector<double, 3> CoordinateSystemTrafo::transformFrom(const ippl::Vector<double, 3>&)`
- `ippl::Vector<double, 3> CoordinateSystemTrafo::transformTo(const ippl::Vector<double, 3>&)`
- `ippl::Vector<double, 3> CoordinateSystemTrafo::rotateTo(const ippl::Vector<double, 3>&)`
- `ippl::Vector<double, 3> CoordinateSystemTrafo::rotateFrom(const ippl::Vector<double, 3>&)`

They are used by the OrbitThreader to track the reference particle (on host).
To transform the bunch during the tracking the we now have the functions:

- `CoordinateSystemTrafo::transformBunchTo(ViewType Rview)`
- `CoordinateSystemTrafo::transformBunchFrom(ViewType Rview)`
- `CoordinateSystemTrafo::rotateBunchTo(ViewType Pview)`
- `CoordinateSystemTrafo::rotateBunchFrom(ViewType Pview)`

They launch a kernel and copy all the necessary information (rotation matrix, origin) to device. For example:
```
template <typename ViewType>
inline void CoordinateSystemTrafo::transformBunchTo(ViewType Rview) const {
    
    // Information to copy
    OpalMatrix_t rot = rotationMatrix_m;
    ippl::Vector<double, 3> ori = origin_m;
    
    // Kernel launch over all particles
    Kokkos::parallel_for("transformBunchTo", Rview.extent(0), 
        KOKKOS_LAMBDA(const size_t i) {
            
            // Apply transformation to all particles
            ippl::Vector<double, 3> delta = Rview(i) - ori;
            Rview(i) = prod_vector(rot, delta);
        });
}
```

## GPU Compatible Beam Line Components
The base class for beam line components has been cleaned up. Each component needs to implement its own 
`Component::apply()` function which applies the E and B field to the particle bunch. 

### Multipole Component
Up till now, the elements generalizes only dipoles and quadrupoles. Extension to higher order magnetic elements is straightforward and limited to the `Multipole::computeField()` and `Multipole::computeFieldHost()` functions.
The implementation produces the same results for GPU (bottom) as for CPU (top). 

image

### DIFF
--- a/src/Algorithms/ParallelTracker.h
+++ b/src/Algorithms/ParallelTracker.h
@@ -1,3 +1,26 @@
+//
+// Class ParallelTracker
+//   OPAL-T tracker.
+//   The visitor class for tracking particles with time as independent
+//   variable.
+//
+// Copyright (c) 200x - 2014, Christof Kraus, Paul Scherrer Institut, Villigen PSI, Switzerland
+//               2015 - 2016, Christof Metzger-Kraus, Helmholtz-Zentrum Berlin, Germany
+//               2017 - 2020, Christof Metzger-Kraus
+//               2025 - present, Ryan Ammann, Paul Scherrer Institut, Villigen PSI, Switzerland
+// All rights reserved
+//
+// This file is part of OPAL.
+//
+// OPAL is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// You should have received a copy of the GNU General Public License
+// along with OPAL. If not, see <https://www.gnu.org/licenses/>.
+//
+
 #ifndef OPAL_ParallelTracker_HH
 #define OPAL_ParallelTracker_HH
 
@@ -39,7 +62,7 @@ class PluginElement;
 /**
  * @brief Implements the simulation loop
  * 
- * ParallelTracker implements the main simulation loop. 
+ * @note ParallelTracker implements the main simulation loop. 
  * TRACK and RUN commands in the inputfile invoke the creation of the 
  * ParalleTracker object and the execution of ParallelTracker::execute()
  */


src/Algorithms/OrbitThreader.cpp
```diff
@@ -470,8 +471,11 @@ double OrbitThreader::computeDriftLengthToBoundingBox(
         || (elements.size() == 1 && (*elements.begin())->getType() == ElementType::DRIFT)) {
         boost::optional<Vector_t<double, 3>> intersectionPoint =
             globalBoundingBox_m.getIntersectionPoint(position, direction);
-        const Vector_t<double, 3> r = intersectionPoint.get() - position;
-        return intersectionPoint ? euclidean_norm(r) : 10.0;
+        if (intersectionPoint) {
+            const Vector_t<double, 3> r = intersectionPoint.get() - position;
+            return euclidean_norm(r);
+        }
+        return 10;
```
aliemen: Where does this "magic number" 10 come from? Could you add a little comment?

src/Utilities/MSLang/AffineTransformation.h
```diff
@@ -1,7 +1,8 @@
 #ifndef MSLANG_AFFINETRANSFORMATION_H
 #define MSLANG_AFFINETRANSFORMATION_H
 
-#include "Algorithms/BoostMatrix.h"
+//#include "Algorithms/BoostMatrix.h"
+#include <boost/numeric/ublas/matrix.hpp>
```
aaadelmann: Boost still used?

src/Algorithms/OrbitThreader.cpp
rammann: No clue, here I only changed some syntax but no content.

aaadelmann (approved):

PR #170, opened 2026-01-17
Implements OPALX_FIELD_DEBUG into new cmake

### LINKED ISSUE #169: OPALX_FIELD_DEBUG is missing from cmake
Writing the fields to a file every timestep is [already implemented](https://github.com/OPALX-project/OPALX/blob/9a886cdeadd71b47795dec458bf7988e26495b1d/src/PartBunch/FieldSolver.cpp#L382). However, it looks like, this flag does not exist in the cmake yet.

I fixed this already in my branch, but thought, it might make sense to cherry-pick it into a small merge.

aliemen:
closes #169 

Adds a new flag to enable the field output in the fieldsolver routine:
```bash
cmake ... -DOPALX_FIELD_DEBUG=ON ...
```

### DIFF
--- a/src/PartBunch/FieldSolver.cpp
+++ b/src/PartBunch/FieldSolver.cpp
@@ -381,14 +381,12 @@ void FieldSolver<double,3>::runSolver() {
             if constexpr (Dim == 2 || Dim == 3) {
 #ifdef OPALX_FIELD_DEBUG
                 this->dumpScalField("rho");
-                // call_counter_m++;
 #endif
 
                 std::get<FFTSolver_t<double, 3>>(this->getSolver()).solve();
 #ifdef OPALX_FIELD_DEBUG
                 this->dumpScalField("phi");
                 this->dumpVectField("ef");
-                call_counter_m++;
 #endif
             }
         } else if (this->getStype() == "P3M") {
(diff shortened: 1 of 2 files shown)


aliemen:
Thanks @mohsensadr! Also if you already looked at it, feel free to accept the request.

I compiled and tested it on Mac and it works as intended. The info message is printed during the cmake call and fields are written to a file after every timestep. 

Only the `callcounter_m` is incremented twice every timestep, so the naming of the files is a bit unintuitive. I'll comment out the second increment in a bit.

src/PartBunch/FieldSolver.cpp
```diff
@@ -381,7 +381,7 @@ void FieldSolver<double,3>::runSolver() {
             if constexpr (Dim == 2 || Dim == 3) {
 #ifdef OPALX_FIELD_DEBUG
                 this->dumpScalField("rho");
-                call_counter_m++;
+                // call_counter_m++;
```
mohsensadr: why don't you remove the line instead of commenting it out

`// call_counter_m++;`

src/PartBunch/FieldSolver.cpp
```diff
@@ -399,7 +399,7 @@ void FieldSolver<double,3>::runSolver() {
             if constexpr (Dim == 3) {
 #ifdef OPALX_FIELD_DEBUG
                 this->dumpScalField("rho");
-                call_counter_m++;
+                // call_counter_m++;
```
mohsensadr: same here

src/PartBunch/FieldSolver.cpp
aliemen: I didn't remove it, because I wasn't sure why exactly it was added. For example, it could be a call counter that counts the number of `this->dump[...]Field(...)` calls and it is only incremented inside the `OPALX_FIELD_DEBUG` blocks (I guess to have a running number for field output files, but then it would need to be incremented three times per call). 

However, I guess this doesn't make too much sense and since it's a member variable of `FieldSolver`, I'll just "define" it as number of solver calls. In that case, I can remove it and have just one single increment in the `runSolver` function.

mohsensadr: sounds good

mohsensadr (approved):

PR #195, opened 2026-02-05
Added an integral feature to GSLCubicSpline

Jonathan-A-Thompson:
Added an integral feature to GSLCubicSpline along with the GSL compatible interface and a test case.  
Also had to add #include <numbers> to matheval.hpp to allow it to build with Apple Clang, hope that's ok.

### DIFF
--- a/src/Utilities/GSLCubicSpline.h
+++ b/src/Utilities/GSLCubicSpline.h
@@ -218,8 +218,9 @@ private:
 
     /// \brief Compute the integrals of the intervals.
     void computeIntegrals() {
+        integrals_.resize(x_.size() - 1);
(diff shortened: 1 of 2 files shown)


src/Utilities/GSLCubicSpline.h
```diff
@@ -162,7 +215,41 @@ class CubicSpline {
         b_[n - 1] = 0.0;
         d_[n - 1] = 0.0;
     }
-    
+
+    /// \brief Compute the integrals of the intervals.
+    void computeIntegrals() {
```
mohsensadr: shouldn't we clear `integrals_` before the for loop?

PR #199, opened 2026-02-10
196 port time dependency classes

Jonathan-A-Thompson:
There were three time dependency classes in old Opal:  SplineTimeDependence, SinusoidalTimeDependence and PolynomialTimeDependence.  They are all derived from the AbstractTimeDependence class.  The idea is that they can be used to control time varying aspects of Opal elements.  As I understand things at the moment, an element calculates the E and B fields for a block of particles at a specific time and this will be done on CPU & GPU devices (for example, Multipole has the computeField kernel function).  The time dependency value would be a constant input to this calculation and can therefore be performed once on the host for all the instances of the kernel.  This fits well with the host only implementation of the cubic spline that the SplineTimeDependence class requires.  If I have misunderstood any of this, please let me know.

Old versions of two of the three time dependence classes were already present in OpalX.  I have ported the latest versions of all three into src/Algorithms and their associated UI classes into src/Elements.  I have also created a new directory unit_tests/Algorithms and placed their unit tests there.  I followed how they are named and utilised in the other directories of unit_tests, hope that's ok.

While I was doing this I made a few c++20-ish changes like marking functions with override, using for-all style loops, etc.  What I haven't done yet is change SplineTimeDependence to use the new cubic spline API, it still is using the GSL interface functions.

### DIFF
--- a/src/Algorithms/ParallelTracker.cpp
+++ b/src/Algorithms/ParallelTracker.cpp
@@ -713,7 +713,9 @@ void ParallelTracker::pushParticles(const BorisPusher& pusher) {
     itsBunch_m->switchOffUnitlessPositions(true);
     /// \todo update gives different results on one rank?
     //itsBunch_m->getParticleContainer()->update();
+    Kokkos::fence();
     ippl::Comm->barrier();
+    itsBunch_m->bunchUpdate();
 }
 
 /**
--- a/src/PartBunch/PartBunch.cpp
+++ b/src/PartBunch/PartBunch.cpp
@@ -167,8 +167,10 @@ void  PartBunch<T, Dim>::gatherLoadBalanceStatistics() {
 
 template <typename T, unsigned Dim>
 void PartBunch<T, Dim>::setSolver(std::string solver) {
+    Inform m("PartBunch::setSolver");
+    m << "Initializing solver: " << solver << endl;
     if (this->solver_m != "")
-        *gmsg << "* Warning solver already initiated but overwrite ..." << endl;
+        m << "Warning solver already initiated but overwrite ..." << endl;
 
     this->solver_m = solver;
 
@@ -181,8 +183,10 @@ void PartBunch<T, Dim>::setSolver(std::string solver) {
         &this->fcontainer_m->getPhi(),
         this->getBCHandler()
     ));
+    m << "Field solver set." << endl;
 
     this->fsolver_m->initSolver();
+    m << "Field solver initialized." << endl;
         
     /// ADA we need to be able to set a load balancer when not having a field solver
     this->setLoadBalancer(std::make_shared<LoadBalancer_t>(
@@ -191,8 +195,7 @@ void PartBunch<T, Dim>::setSolver(std::string solver) {
         this->pcontainer_m, 
         this->fsolver_m
     ));
-    
-    *gmsg << "* Solver and Load Balancer set" << endl;
+    m << "Solver and Load Balancer set." << endl;
 }
 
 template <typename T, unsigned Dim>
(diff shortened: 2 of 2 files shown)


aaadelmann:
It would be nice to understand if the the relation with the reimplemented GSL functionality of Mohsen and the one you implemented.

src/Algorithms/PolynomialTimeDependence.cpp
```diff
@@ -26,15 +26,43 @@
  */
 
 #include "Algorithms/PolynomialTimeDependence.h"
-
 #include "Utility/Inform.h"
+#include <iostream>
+
+PolynomialTimeDependence* PolynomialTimeDependence::clone() {
+    const auto temp(coeffs);
+    auto* d = new PolynomialTimeDependence(temp);
```
aaadelmann: maybe adopt smart pointers here

Jonathan-A-Thompson:
> It would be nice to understand if the the relation with the reimplemented GSL functionality of Mohsen and the one you implemented.

The original GSL spline functions supported both linear and cubic interpolation between data points.  The GSLCubicSpline class implemented by Mohsen only does cubic interpolation (as implied by the class name).  The SplineTimeDependence class I have ported from old Opal requires both cubic and linear interpolation.  I have therefore implemented linear interpolation in a second class GSLLinearSpline following the same API as Mohsen's class as I felt it was too big a change to make to his class.  However, I think it would be an improvement to provide a base class that contained the common code to both and defined the API.   I'm also thinking the GSL prefix on the class names should be dropped and the GSL compatible interface moved into its own file.   I am happy to make some or all of those changes if you guys approve.

src/Utilities/LinearSpline.h
```diff
@@ -0,0 +1,83 @@
+//
+// Cubic Spline Interpolation to replace GSL spline
```
mohsensadr: same here.

src/Elements/OpalSinusoidalTimeDependence.cpp
```diff
@@ -0,0 +1,63 @@
...
+    "frequency, amplitude, phase offset, DC offset, given by "
+    "f(t) = sigma_over_i(a[i] / 2 * sin(2 * pi * f[i] * t + p[i]) + o[i])";
+
+OpalSinusoidalTimeDependence::OpalSinusoidalTimeDependence()
+    : OpalElement(static_cast<int>(SIZE), "SINUSOIDAL_TIME_DEPENDENCE", doc_string.c_str()) {
+    itsAttr[FREQUENCIES] = Attributes::makeRealArray(
+        "FREQUENCIES",
+        "Sine wave frequencies, length determines the number of size waves included.");
+    itsAttr[AMPLITUDES] = Attributes::makeRealArray(
+        "AMPLITUDES", "Peak-to-peak amplitude for each size wave.  If undefined, defaults to 1.0.");
```
mohsensadr: same here

src/Utilities/LinearSpline.h
Jonathan-A-Thompson: Again

src/Elements/OpalSinusoidalTimeDependence.cpp
```diff
@@ -0,0 +1,63 @@
...
+    : OpalElement(name, parent) {
+}
+
+void OpalSinusoidalTimeDependence::update() {
+    AbstractTimeDependence::setTimeDependence(
+        getOpalName(), std::make_shared<SinusoidalTimeDependence>(
+                           Attributes::getRealArray(itsAttr[FREQUENCIES]),
+                           Attributes::getRealArray(itsAttr[PHASE_OFFSETS]),
+                           Attributes::getRealArray(itsAttr[AMPLITUDES]),
+                           Attributes::getRealArray(itsAttr[DC_OFFSETS])));
```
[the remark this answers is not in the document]
Jonathan-A-Thompson: Ok, swapped around.

mohsensadr:
@Jonathan-A-Thompson please feel free to create another PR once you fixed the issue.

PR #214, opened 2026-02-12
fix licences close #206

### LINKED ISSUE #206: License files need to be updated
The main license file is a [symlink](https://github.com/OPALX-project/OPALX/blob/master/LICENSE) that does not work in GitHub, we should probably link a correct license file for GitHub to recognize. 

Next, the [`Licenses/`](https://github.com/OPALX-project/OPALX/tree/master/LICENSES) directory contains many unnecessary files that can be removed and/or modernized. The general license structure should at some point probably be reviewed/cleaned up.

This is just a general issue and does not press for immediate action. However, there is no "OPALX" license, only a "OPAL" license and countless files in the repository still refer to "OPAL" instead of "OPALX".

aaadelmann:
fix licences close #206


mohsensadr (changes requested):
As mentioned in the conversation, I think LICENSE in the root of repo should point to:
LICENSE -> LICENSES/OPALX/LICENSE

aliemen:
Honestly, I don't see a reason why we would even want to keep the LICENSES directory. We never actually _use_ (in the sense of copying their files into our source repo) any of these codes we provide licenses for anyways, so there is no need to provide them. The only license we need is the one for OPALX itself. 

Note that basically all codes do the same (have only one LICENSE): [IPPL](https://github.com/IPPL-framework/ippl), [heFFTe](https://github.com/af-ayala/heffte), [gromacs](https://github.com/gromacs/gromacs), [epoch](https://github.com/epochpic/epoch), [kokkos](https://github.com/kokkos/kokkos/blob/develop/LICENSE) and every other big repository I could find. 

That's why I propose my latest commit (obviously feel free to revert/reset it should you disagree).

mohsensadr:
My suggestion is to discuss this with TTO of PSI. We are publishing a code that PSI owns its IP, and the code relies on open source dependencies that have their own licenses. Like @aliemen, I don't think we need to include their license here (licenses of dependencies are included in _dep directory when opalx is built from scratch anyways) but I'm not an expert of IP or license (which can get very complicated).

If we are sued for any IP related reasons, PSI's TTO would have to defend the case. So, it makes sense to get their agreement as soon as possible (I guess Andy may have done it before with the old OPAL, we just need to renew it).

aaadelmann:
First in case we are using external SW we are not totally free to chose the licence. This was checked years ago for the old OPAL. I suggest we are not changing that for the moment.

PR #219, opened 2026-02-13
Add cmake compile time flag to indicate device backend + create HIP/GFX90A runner

### LINKED ISSUE #146: Compilation issue for backends other than CUDA
The following code does not compile on Lumi (or any environment other than CUDA for that matter):
https://github.com/OPALX-project/OPALX/blob/28804c3089c8e9cabcc3532034460bcd24346f21/src/PartBunch/Binning/ParallelReduceTools.h#L232-L292

A quick fix is simply using different flags instead of `KOKKOS_ENABLE_CUDA`, like `Kokkos_ENABLE_HIP` (AMD) or `Kokkos_ENABLE_SYCL` (Intel).

### LINKED ISSUE #215: Parallel Reduction Tools need an AMD branch
Parallel Reduction Tools need AMD branch to avoid compiling issues

aliemen:
closes #215 
closes #216
closes #146 

This pull request introduces a mechanism for detecting device backends (such as CUDA, HIP, SYCL) and defines a general compilation flag (`OPALX_DEVICE_COMPILATION`) for device-specific code. It also moves the `cmake/Messages` include a bit further up in the cmake hirachy to allow colored messages everywhere.

The idea is to allow backend dependend code [like here](https://github.com/OPALX-project/OPALX/blob/8dc1cdd3b0ef2350595870b6d63e9a1d24227fb4/src/PartBunch/Binning/ParallelReduceTools.h#L232) to be compiled for all device backends without needing to hardocde CUDA/HIP/...

The new cmake function will check if `PLATFORMS` contains a backend that is not on host, [see Kokkos docs](https://kokkos.org/kokkos-core-wiki/get-started/configuration-guide.html#serial-backend), and define the corresponding flag that can be used in the code.

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -12,7 +12,7 @@ jobs:
   # ===== CPU SERIAL BUILD =====
   build-cpu-serial-debug:
     runs-on: ubuntu-latest
-    if: contains(github.event.pull_request.labels.*.name, 'compile-ci') || github.event.pull_request.draft == false
+    if: contains(github.event.pull_request.labels.*.name, 'compile-ci') && github.event.pull_request.draft == false
           
     steps:
       - name: Checkout code
@@ -52,6 +52,7 @@ jobs:
   # ===== CPU OPENMP BUILD =====
   build-cpu-openmp-debug:
     runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'compile-ci') && github.event.pull_request.draft == false
 
     steps:
       - name: Checkout code
@@ -114,7 +115,8 @@ jobs:
   # ===== GPU CUDA BUILD =====
   build-cuda-debug:
     runs-on: ubuntu-latest
-    if: contains(github.event.pull_request.labels.*.name, 'compile-ci') || github.event.pull_request.draft == false
+    if: contains(github.event.pull_request.labels.*.name, 'compile-ci') && github.event.pull_request.draft == false
+    
     container:
       image: nvidia/cuda:13.1.0-devel-ubuntu24.04
     steps:
@@ -162,9 +164,10 @@ jobs:
   # ===== GPU HIP BUILD =====
   build-hip-debug:
     runs-on: ubuntu-latest
-    if: contains(github.event.pull_request.labels.*.name, 'compile-ci') || github.event.pull_request.draft == false
+    if: contains(github.event.pull_request.labels.*.name, 'compile-ci') && github.event.pull_request.draft == false
+    
     container:
-      image: rocm/dev-ubuntu-22.04:latest
+      image: rocm/dev-ubuntu-22.04:7.2
     steps:
       - name: Checkout code
         uses: actions/checkout@v3


aliemen:
I added a new runner that compiles OPALX for AMD_GFX90A architecture with HIP. On my test it [ran fine](https://github.com/OPALX-project/OPALX/actions/runs/21992699811/job/63543974533?pr=220), so I merged it into this branch. If it compiles on all the hardware, then the cmake change should also be fine for LUMI.

aliemen:
@mohsensadr thanks for looking at it. I *think* that standard runners are actually free for us? We already used $\sim 3800$ minutes last month without a problem, right? At least [here](https://github.com/orgs/community/discussions/156389#discussioncomment-12807955) they say that they should be free and unlimited with a maximum of [20 concurrent runners](https://docs.github.com/en/actions/reference/limits#job-concurrency-limits-for-github-hosted-runners).

aliemen:
OK, I just tested everything I wrote in the previous comment, seems to work fine 👍🏻. 

@aaadelmann let me know if you would change anything. I think apart from that you would only need to edit the "*Require status checks to pass before merging*" setting in the branch protection rules.

.github/workflows/workflow.yml
```diff
@@ -5,12 +5,14 @@ on:
...
   pull_request:
+    types: [opened, synchronize, reopened, labeled, ready_for_review]
     branches: [ "master" ]
           
 jobs:     
   # ===== CPU SERIAL BUILD =====
-  build-cpu-serial-debug:  
+  build-cpu-serial-debug:
     runs-on: ubuntu-latest
+    if: contains(github.event.pull_request.labels.*.name, 'compile-ci') || github.event.pull_request.draft == false
```
mohsensadr: shouldn't or || be and && instead ?

Run the job only if
• The PR has the compile-ci label
AND
• The PR is not a draft

CMakeLists.txt
```diff
@@ -24,13 +24,18 @@ project(OPALX VERSION 1.0 LANGUAGES C CXX)
...
+include(Messages)
+
 # ------------------------------------------------------------------------------
 # Primary OPALX options
 # ------------------------------------------------------------------------------
 include(OPALXOptions)
 
 # ------------------------------------------------------------------------------
-# Project setup
+# Project setup (call first)
```
mohsensadr: what do you mean by "(call first)"?

src/PartBunch/Binning/ParallelReduceTools.h
```diff
@@ -279,14 +279,16 @@ namespace ParticleBinning {
...
-         * @note This constructor is not intended to be used on CUDA, as `HostArrayReduction` is not supported there.
-         *       Instead, it throws an error to indicate that this functionality is not available.
+         * @note This constructor is not intended to be used on device backends (CUDA/HIP/SYCL),
+         *       as `HostArrayReduction` is not supported there. Instead, it throws an error
+         *       to indicate that this functionality is not available.
          */
         KOKKOS_INLINE_FUNCTION
         HostArrayReduction& operator+=(const HostArrayReduction& src) {
-            Kokkos::abort("Error: HostArrayReduction is not supported on CUDA!\n       Note: It exists only for compilation compatibility.");
+            (void)src;  // Silence unused parameter warning on Clang
```
mohsensadr: I don't think it is a good idea to globally silence warnings like this, even if it is for unused parameter. The warning should be resolved in the code. please let me know what you think.

mohsensadr (changes requested):

.github/workflows/workflow.yml
```diff
@@ -99,7 +101,7 @@ jobs:
       #       --exclude '.*_deps.*' \
       #       --exclude '.*3p.*' \
       #       --exclude '.*unit_tests.*' \
-      #       --gcov-ignore-errors=all \
+      #       --gcov-ignore-errors=all \
```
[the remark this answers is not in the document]
aliemen: Sorry, I added those to have something to commit for the CI tests. Will remove

PR #226, opened 2026-02-17
196 port time dependency classes

Jonathan-A-Thompson:
Another go at merging the time dependency classes.  I have also added a test case for one of the manual tests that was mentioned.


Jonathan-A-Thompson:
> Just to clarify, besides the "compile-ci" label, the PR should be also "Ready for review" and not "draft" for CI to run.

Yes, understood.  I switched it to draft when I discovered a merge from main had thrown away all my changes.  I didn't have time to recover that yesterday.

Jonathan-A-Thompson:
As requested I have removed the Sanity tests.

mohsensadr:
Thank you @Jonathan-A-Thompson. Since you don't have access to GPUs, I can test it on a local GPU before approving it.
For the regression tests of opalx (which is in the making), please see:
https://amas.pages.psi.ch/opal-live-doc/opalx-regression-test/overview/master/index.html

src/Elements/CMakeLists.txt
```diff
@@ -35,6 +39,10 @@ set (HDRS
     OpalOffset/OpalLocalCylindricalOffset.h
     OpalVerticalFFAMagnet.h
     OpalProbe.h
+    OpalPolynomialTimeDependence.h
+    OpalSplineTimeDependence.h
+    OpalSinusoidalTimeDependence.h
+
```
mohsensadr: please remove the extra line.

Jonathan-A-Thompson:
> Do we have a input file for a Regression test?

There's no need.  I was just trying to include a sanity test.  I'm guessing your regression tests will do the job as they currently stand.

PR #248, opened 2026-02-26
Add clang workflow check for changed files

### LINKED ISSUE #247: Add clang formatting
With potentially more people working on it and to preserve code quality, we should do what IPPL does and use clang-formatting to check our source code.

This is a quite easy task, so I implemented it quickly (just as a test).

aliemen:
I ran the following two commands on the codebase:
```
find src/ -type f \( -name "*.h" -o -name "*.cpp" -o -name "*.hpp" -o -name "*.tpp" \) -exec clang-format -i -style=file {} +
find unit_tests/ -type f \( -name "*.h" -o -name "*.cpp" -o -name "*.hpp" -o -name "*.tpp" \) -exec clang-format -i -style=file {} +
```
These commands applied our clan-format to the whole code base, which is an insanely huge commit. Alternatively, I could also revert this commit and only add the new workflow file.

The new workflow checks modified files and fails when the code does not comply with our clang-format rules. For the new workflow, I more or less copied what was [done in IPPL](https://github.com/IPPL-framework/ippl/blob/master/.github/workflows/clang-format.yml). I use `clang-format-20` here. But more specifically `20.1.8` through a python wrapper, since this allows using auto-clang formatting also on Mac.


aliemen:
@aaadelmann not quite. What I did here is simply a "check" that tells you "_format your code correctly (clang-format-20.1.8) or the workflow fails_". 

IPPL does have the same already, though. See [here](https://github.com/IPPL-framework/ippl/blob/master/.github/workflows/clang-format.yml). This new PR you linked seems to give the user the option to install clang-format and have it format all of your changed files automatically when you do the commit. This, however, is from the user perspective completely optional. But I guess, should we proceed with my clang-check, we could just copy the hook scripts from IPPL. The drawback is that you still need to install clang-format locally for this pre commit hook.

aliemen:
OK you're right, since we then probably just copy the workflow file from IPPL, this PR is obsolet. Feel free to close

aaadelmann (approved):

PR #273, opened 2026-03-10
fix TestBinning compilation error on Gwendolen 

aaadelmann:
fix compilation error TestBinning.cpp using a functor outside of GTest function


aliemen:
Hm interesting, putting the initialization in a separate functor changed the results. I will fix it today or tomorrow as soon as I have a few minutes.

Issue #257, opened 2026-03-03
Implement Standing Wave RF Cavity in OPALX


MichZampetakis:
OPALX currently lacks RF cavity elements, which are required to model SwissFEL for the Dense Coulomb Clouds project. The existing `RFCavity` in OPAL provides a reference implementation but is CPU-only and tied to legacy code.

The goal of this issue is to revise the OPAL standing wave RF cavity into a GPU-compatible class for OPALX. 
_Travelling wave cavities will be implemented in a future, separate issue._

PR #302, opened 2026-03-20
Migration of Standing Wave RF Cavity from OPAL to OPAL-X #257

MichZampetakis:
### **Description:**

This PR migrates the **Standing Wave RF Cavity** implementation from OPAL to OPAL-X using **Kokkos** for portability.

**Modified files:**

* `src/AbsBeamline/RFCavity.cpp`
* `src/AbsBeamline/RFCavity.h`

To support running test cases using legacy **AWA Gun fieldmaps**, the following were also migrated:

* `src/Fields/FM2DDynamic.cpp`
* `src/Fields/FM2DDynamic.h`

Additionally, `FM2DDynamic` is now included in `src/Fields/Fieldmap.cpp` to integrate properly with the field infrastructure.

**Testing:**

* Verified against OPAL in serial and OpenMP mode, as well as on GPU (GH200). 
* Unit tests:
     - `FM2DDynamic` fieldmap
     - `RFCavity` element (planned, seperate issue will be opened)
* Regression tests: 
     - Planned in separate repository. https://github.com/OPALX-project/regression-tests-x/issues/12

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -8,6 +8,15 @@ on:
     types: [opened, synchronize, reopened, labeled, ready_for_review]
     branches: [ "master" ]
 
+# Cancel any in-progress runs of this workflow for the same pull 
+# request when a new commit is pushed or when the PR is updated 
+# with new labels. This ensures that only the latest run for a 
+# given PR is active, preventing redundant builds and saving CI 
+# resources.
+concurrency:
+  group: ${{ github.workflow }}-pr-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
 jobs:
   # ===== CPU SERIAL BUILD =====
   build-cpu-serial-debug:


aliemen:
Just tried reproducing it (on GH200 as well as OPENMP), and I think I more or less was able to. I used your provided input script and ran it both with OPAL and OPALX. This comment only looks at the results, I will review the code again later, but at a glance it looked OK. You only need to update the branch with master: _but be careful, you might have to change the `applyField` signature of your `2DDynamic` field map_! (since I had to change it for the `Solenoid` scaling, see #336) 

## Energy 

Energy looks perfectly fine to me. OPALX in openmp/GH200 is consistent with my old OPAL run:

zenergy

## RMS Momenta and Positions

Looks really good in the beginning, but then has this weird peak exactly where the cavity ends. Note that the peak only appears in my old OPAL run:

zrms-mom

RMS positions are similar. The beginning looks perfect. Then, at the end we see how increasing/and then peaking momenta lead to a spread in the bunch:

zrms-pos

## Final remarks

I think the energy spread should be ignored. I am pretty sure that there has to be another unit inconsistency between OPAL/OPALX:

zenergy-spread

Apart from that it looks good to me. I have to admit that I don't understand the "peak" where the element ends (looking at RMS momenta), but perhaps that's just some kind of de-phasing? Do you know more about this @MichZampetakis? Or perhaps there is a difference in how the cavity is applied compared to old OPAL. This simulation uses a quite big spread in `s` direction, so there might be a point where half the bunch is in the cavity while the other half isn't. I am pretty sure that this is also the reason why we see some "movement" in the bunch after the cavity ended (since "`s`" is only a mean quantity).

aliemen:
Looks good apart from my comment. Merge master and I'll take a final look/approve it (_I was also able to more or less reproduce full `AWAGun-1`, but that will be in my next PR_). 

If you didn't read my comment above: merging master might produce an error; you need to make sure that the `applyField` signature is correct. Something like
```cpp
void applyField(std::shared_ptr<ParticleContainer_t> pc, double) override;
```
...instead of "`applyField(... pc)`" should do the trick.

PR #309, opened 2026-03-23
H5Part:: switch from cscs to github

aaadelmann:
switch from cscs to github


aliemen (approved):
I still saw a lot of H5hut warnings (all of them were `sign-compare` and `unused-parameter`), but if that's the new official repo, then it makes sense.

Also: ssh access doesn't work when you don't have your keys set up in github. Since we fetch ressources from a public repo, I changed it to https. From my side feel free to merge

PR #312, opened 2026-03-24
Implement `deleteParticlesOutside` function and deletes particles if necessary

### LINKED ISSUE #311: `BOUNDPDESTROYFQ` is not implemented
Currently, particles are never deleted in OPALX. However, if I want to test space charge, it would be nice to have this feature. Here is why:

A flattop emission without space charge can look like this after some timesteps:

Image

Activating space charge then looks like this (same scale as before, the "tile" in the front is slightly bigger):

Image

However, with space charge we see the following: 

https://github.com/user-attachments/assets/0250d107-3c2e-4d38-85e0-7c713fb56e82

Note that the little "line" in the middle now corresponds to the "flattop line" in the previous images, meaning:

Image

One can see that there are always a few particles that "explode away". These lead to the domain being way too big, which in turn leads to very inaccurate self fields, since 99% of bunch are located within "one grid cell". The solution is to delete these escaping particles (they don't contribute to any bunch effects anyways), similar to old OPAL. 

---

This shouldn't be a big task, since the functionality already exists in IPPL, so I'll just go ahead and write a function in `ParallelTracker` that checks if particles are "`BOUNDPDESTROYFQ`" many sigmas outside of the mean and send them to the IPPL delete particles function. Shouldn't affect anything else in OPALX, so no big refactoring necessary.

aliemen:
closes #311

Call IPPLs `destroy` on particles that are outside "`Options::boundpDestroyFreq`" many sigmas of the mean of the bunch. Works perfectly fine on CPU:

https://github.com/user-attachments/assets/0c2c838c-2330-4693-bb3c-d679dfc611b1

This is the equivalent to the video shown in #311, just a bit zoomed in and slowed down with bigger ParaView dots (so one can see the individual particles that are escaping). One can see that the escaping particles are deleted just fine. For a simple one-bin test case, one can see that it makes a difference:

grafik

For a simple flattop emission with a very big rms, we wouldn't expect such an explosion of rms - so this makes way more sense (I used rms in x and y of `0.01m` with `1nC`).

---

Note that I expect some of the regression tests to fail from this change, since space charge effect are now potentially handled with much more resolution (for some cases like the flattop from above). 
[Here](https://github.com/user-attachments/files/26218047/Drift-4-open-flattop-bins.txt) the adapted input file that you can use to reproduce the example above.

### DIFF
--- a/src/AbsBeamline/ConstantEFieldCavity.cpp
+++ b/src/AbsBeamline/ConstantEFieldCavity.cpp
@@ -80,8 +80,7 @@ bool ConstantEFieldCavity::apply() {
     const double Ey         = Ey_m;
     const double Ez         = Ez_m;
 
-    Kokkos::parallel_for(
-        "ConstantEFieldCavity::apply()", Kokkos::RangePolicy<>(0, nLocal),
+    Kokkos::parallel_for("ConstantEFieldCavity::apply()", nLocal,
         KOKKOS_LAMBDA(const size_t i) {
             if (Rview(i)[2] > 0.0 && Rview(i)[2] <= elemLength) {
                 Eview(i)[0] += Ex;
--- a/src/AbsBeamline/Multipole.cpp
+++ b/src/AbsBeamline/Multipole.cpp
@@ -278,21 +278,19 @@ bool Multipole::apply()
     int maxSkew = max_SkewComponent_m;
 
     // Kernel launch over all particles
-    Kokkos::parallel_for("Multipole::apply()", Kokkos::RangePolicy<>(0, nLocal), 
-    KOKKOS_LAMBDA(const int i)
-    {
-        // Check bounds
-        if (Rview(i)(2) > 0 && Rview(i)(2) <= elemLength){
-            Vector_t<double,3> Ef(0.0), Bf(0.0);
-            // Compute field at particle position
-            computeField(Rview(i), Ef, Bf, 
-                normalComponents, skewComponents,maxNormal, maxSkew);
-            for(unsigned d=0; d<3; ++d){
-                Eview(i)(d) += Ef(d);
-                Bview(i)(d) += Bf(d);
+    Kokkos::parallel_for("Multipole::apply()", nLocal, 
+        KOKKOS_LAMBDA(const size_t i) {
+            // Check bounds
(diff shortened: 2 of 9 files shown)


aliemen:
Looks good on A100:

https://github.com/user-attachments/assets/63de1d97-e3df-4297-89a1-55760e0ee842

Also runs fine multirank and on CPU/GPU it shows the expected deletions that were outlined in #311:

```text
OPAL-X> * 16:29:23 Step      7 at 46.764 [mm], t= 800.000 [ps], E=26.924 [keV]
OPAL-X[2]> * Wrote beam statistics.
OPAL-X> * 16:29:23 Step      8 at 52.610 [mm], t= 900.000 [ps], E=24.890 [keV]
OPAL-X[2]> * Wrote beam statistics.
ParallelTracker::execute[2]> Deleted 1 particles outside 10-sigma boundary, 3047 remaining.
OPAL-X> * 16:29:23 Step      9 at 58.455 [mm], t= 1.000 [ns], E=23.293 [keV]
OPAL-X[2]> * Wrote beam statistics.
ParallelTracker::execute[2]> Deleted 1 particles outside 10-sigma boundary, 3373 remaining.
OPAL-X> * 16:29:23 Step     10 at 64.301 [mm], t= 1.100 [ns], E=21.508 [keV]
OPAL-X[2]> * Wrote beam statistics.
ParallelTracker::execute[2]> Deleted 7 particles outside 10-sigma boundary, 3693 remaining.
OPAL-X> * 16:29:24 Step     11 at 70.146 [mm], t= 1.200 [ns], E=20.097 [keV]
OPAL-X[2]> * Wrote beam statistics.
ParallelTracker::execute[2]> Deleted 13 particles outside 10-sigma boundary, 4007 remaining.
OPAL-X> * 16:29:24 Step     12 at 75.992 [mm], t= 1.300 [ns], E=17.766 [keV]
OPAL-X[2]> * Wrote beam statistics.
ParallelTracker::execute[2]> Deleted 28 particles outside 10-sigma boundary, 4306 remaining.
OPAL-X> * 16:29:24 Step     13 at 81.837 [mm], t= 1.400 [ns], E=15.146 [keV]
OPAL-X[2]> * Wrote beam statistics.
ParallelTracker::execute[2]> Deleted 31 particles outside 10-sigma boundary, 4602 remaining.
...
```

My local regression tests look good.

PR #324, opened 2026-03-31
Unify bunch states with new `BunchStateHandler` class

### LINKED ISSUE #323: "Bunch states" need unification
OPALX uses quite a few flags in the code that indicate the state of the simulation. Since we might need more in the future (talking about #300 or #319 or a `isEmitting` flag for emission processes), it makes sense to have some kind of unification for these flags. There are two possibilities:
- Have a public namespace with some variables/bools/enums that can be accessed and read/set from everywhere (easy).
- Create a `BunchStateHandler` class that keeps track of all of these flags and provides access through getter/setter/access functions.

I favor the second option, since it gives much more freedom in terms of checking or tracking state changes. This handler can be part of the `ParticleContainer` and accessed from there (where it is ultimately saved should be easy to change, though). A lot of flags, like `isUnitless` or (tbd) `momentsDirty` are probably species specific, so it would be nice to have the option of multiple particle containers.

aliemen:
closes #323 

(the big diff in `DistributionMoments` comes from clang-format...)

This PR implements a new `PartBunch/BunchStateHandler.h` class that is supposed to unify all of the states our bunch can be in. Here are the states I found in the codebase that can are currently implemented:
- `unitlessPositions_m`: indicates if the bunch is in unitless positions during any push step.
- `momentsDirty_m`: indicates if `distributionMoments` values are up to date. Resets after every sample, delete, push or kick action.
- `emittingNow_m`: indicates wether any distribuiton is currently considered "emitting". This extends e.g. flattop such that we know when to still apply image charges at the origin (perhaps I'll rename this at some point, since the name might be mistaken for the actual particle emission).
- `firstRepartition_m`: needed later for the loadbalancer, but currently unused.

I completely deleted these flags from e.g. `PartBunch` and used the new handler everywhere necessary. Moments are now short circuited when the `dirty` flag is false. 

--- 

The idea would be that every time a new state variable is added, we can add it to this class. Having a class is in my opinion better than having them defined somewhere globally, since this allows for better "state tracking" should that be necessary. Some states (in the future) might also be a per-bunch action and in that case it would be easy to e.g. give every container their own state handler instance. At the moment, every container (and the `PartBunch`) has a shared pointer to the same state handler instance. Should we want a handler per container, then we would just restrict access to the handler through the container alone. 

---

I also added a unit test file for the new class and integrated the handler into existing unit tests (where necessary). Regression tests pass on CPU and compilation (plus `Drift-4-open-bins`) look good on Merlin7 (GH200). Unit tests pass on CPU and GH200 (Merlin7).

Edit: I know there are ~800 lines changed. But 95% of them are just automatic clang-format on some of the big files (`PartBunch`, `ParallelTracker`, `DistributionMoments`). 

~A fix I put into this PR:~
- ~`BOUNDPDESTROY` will delete all particles if the distribution is exactly 2D (which is the case for FlatTop), so I gave it a little threshold in each dimension.~ No, will be handled in a separate PR #326.


aaadelmann (changes requested):

src/Algorithms/DistributionMoments.cpp
```diff
@@ -119,7 +119,23 @@ void DistributionMoments::computeMoments(
         ippl::ParticleAttrib<Vector_t<double, 3>>::view_type Rview,
         ippl::ParticleAttrib<Vector_t<double, 3>>::view_type Pview,
         ippl::ParticleAttrib<double>::view_type Mview, size_t Np, size_t Nlocal) {
-    Np = (Np == 0) ? 1 : Np;
+    // Lock the per-container slot. Null = never bound; expired = owning
```
aaadelmann, line 122: I do not understand the comment in l122

src/Algorithms/DistributionMoments.cpp
```diff
@@ -222,6 +238,11 @@ void DistributionMoments::computeMoments(
 
     double betaGamma = std::sqrt(std::pow(meanGamma_m, 2) - 1.0);
     geometricEps_m   = normalizedEps_m / Vector_t<double, 3>(betaGamma);
+
+    // Cache is now consistent with this container's particle state; clearing
+    // the slot's dirty bit matches the contract advertised in
+    // BunchStateHandler.h. Reuse the strong handle locked at function entry.
+    slot->clearMomentsDirty();
```
aaadelmann: clear means moments are not Dirty i.e. clean? maybe reconsider wording.

src/PartBunch/BunchStateHandler.h
```diff
@@ -0,0 +1,114 @@
...
+     * helper is stateless, so the struct can call it without holding any
+     * handler reference.
+     */
+    struct ContainerState {
+        bool unitlessPositions = false;
+        bool momentsDirty      = true;
+
+        void setUnitlessPositions(bool v);
+        void markMomentsDirty();
+        void clearMomentsDirty();
```
aaadelmann: maybe markMomentsClean as an alternative

aaadelmann (changes requested):

src/PartBunch/BunchStateHandler.h
aliemen: The idea was to keep the variable name "`momentsDirty`" in the getter/setter. But I agree that `markMomentsClean` is more readable. Will change

src/Algorithms/DistributionMoments.cpp
aliemen: A `weak_ptr` does not keep the object alive, it more or less only "observes" a `shared_ptr` that is owned somewhere else. Now to use it, you would "request access" with the "lock" function. 
- If this returns `null`, it means that the `ContainerState` struct was probably never registered via `DistributionMoments::setContainerState`. This is just a safeguard to prevent uninitialzed access.
- If "expired" that would mean the underlying `ContainerState` struct was alrady destroyed (the ParticleContainer stoped existing). In that case, `containerState_m.expired() == true`. But in that case `.lock()` would also return `null`, meaning the `if (!slot)` catches both cases (expired and never set).

aliemen:
Just tested again: Regression and unit tests all pass on CPU + GPU (GH200). Had to fix a bug, since the momentsDirty flag was not marked correctly after update(), but should be all good now.

aaadelmann (approved):

PR #331, opened 2026-04-02
Add PHOTON beam type with validation guardrails

### LINKED ISSUE #330: Add PHOTON as a predefined OPALX BEAM particle with ENERGY-only validation
## Summary

  Add `PHOTON` as an OPALX-only predefined `BEAM::PARTICLE`, with strict validation guardrails and no transport yet.

  This is the first public syntax step toward photon support. The goal is to allow

  ```text
  BEAM, PARTICLE=PHOTON, ENERGY=...

  at the input level, while explicitly rejecting unsupported usage in tracking.

  ## Motivation

  The photon roadmap in the OPALX manual starts with a true PHOTON species and photon-aware particle containers before introducing collision processes, LASERBEAM, luminosity, or Breit-Wheeler physics.

  So the first implementation step should be a safe BEAM-level entry point.

aaadelmann:
## Summary

  Add `PHOTON` as an OPALX-only predefined `BEAM::PARTICLE`, with strict validation guardrails and no transport yet.

  This is the first public syntax step toward photon support. The goal is to allow

  ```text
  BEAM, PARTICLE=PHOTON, ENERGY=...

  at the input level, while explicitly rejecting unsupported usage in tracking.

  ## Motivation

  The photon roadmap in the OPALX manual starts with a true PHOTON species and photon-aware particle containers before introducing collision processes, LASERBEAM, luminosity, or Breit-Wheeler physics.

  So the first implementation step should be a safe BEAM-level entry point.

  ## Scope

  - add PHOTON to the predefined particle infrastructure
  - define PHOTON with
      - mass = 0
      - charge = 0
  - allow BEAM, PARTICLE=PHOTON
  - require ENERGY for photon beams
  - reject photon-beam use of
      - GAMMA
      - PC
      - MASS
      - CHARGE
      - SOURCES
  - reject photon beams in TRACK with a clear exception until photon transport exists
  - update the OPALX manual accordingly

  ## Semantics

  For PARTICLE=PHOTON:

  - ENERGY is the only accepted beam-energy attribute all other attributes are not meaningful
  - the beam is accepted only at definition time
  - TRACK must fail now if a photon beam is selected

  ## Units

  Photon-beam energy should follow the CAIN convention:

  - beam reference energy is specified by energy, not gamma
  - photon BEAM::ENERGY uses eV

  ## Notes

  The manual was also updated.

closes #330


rammann:
I’d like to avoid having `if (isPhoton)` (or similar special-case checks) scattered across multiple call sites.

We could instead keep a single handling point inside `Beam::execute()` or `Beam::update`, where we check for special particles? That way we keep branching logic to a minimum and keep the rest of the Beam pipeline particle-agnostic.

But apart from that it looks good to me! 👍

aaadelmann:
If you refer to the one if in TrackRun, I agree. This will go as soon as we have a more clear idea how to handle photons. The others are ok I think.

PR #340, opened 2026-04-05
Implement `dt` sampling in `FlatTop`

### LINKED ISSUE #339: Implement `dt` sampling in `FlatTop`
Old OPAL samples the emission time directly, while we sample `n` particles every timestep, leading to patterns like this:

Image

I will try to implement fractional timesteps in an attempt to finally get matching `AWAGun` results.

aliemen:
closes #339 

See the issue for a visual explanation of what was changed. This PR samples the "`dt`" attribute uniformaly for all particles for their respective timestep. If `dtG` is the global timestep, then $\Delta t _\mathrm{particle} = \xi \cdot \Delta t _\mathrm{global}$, where $\xi \sim \mathcal{U}(0,1)$. 

It's not the "exact" solution, since this would follow the `FlatTop` distribution itself. However, it's exact for the "flat top" and good enough for the tails if time resolution is good enough.

### DIFF
--- a/src/Distribution/FlatTop.cpp
+++ b/src/Distribution/FlatTop.cpp
@@ -120,12 +120,13 @@ void FlatTop::setInternalVariables(bool emitting,
     sigmaR_m = sigmaR;
 }
 
-void FlatTop::generateUniformDisk(size_type nlocal, size_t nNew) {
+void FlatTop::generateUniformDisk(size_type nlocal, size_t nNew, double dt) {
     if (nNew == 0) { return; }
 
     GeneratorPool rand_pool = rand_pool_m;
     view_type Rview         = pc_m->R.getView();
     view_type Pview         = pc_m->P.getView();
+    auto dtView             = pc_m->dt.getView();
 
     double pi                  = Physics::pi;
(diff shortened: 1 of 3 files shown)


aliemen:
Here is an animation of the bunch with and without the fractional timesteps. Here is `AWAGun-1` without the fractional timesteps:


https://github.com/user-attachments/assets/b3c0b87e-bcd1-4955-aff5-060a1315c72a

And here the exact same with fractional timesteps (smooth distribution in `z` direction) in the exact same scale:


https://github.com/user-attachments/assets/a0b7f4ff-6749-4b7f-8933-68353e4131f7

The difference here is that having these individual "discs". Here you see the exact same input file (and same scale) for old OPAL:


https://github.com/user-attachments/assets/e8d3e2e8-9d90-4357-9d94-b84dadb253ae

As you can see, there are still some differences, but this weird "disc" that separates because of [these](https://github.com/OPALX-project/OPALX/issues/339#issue-4207265653) separations in the sampling are gone. 

The difference in shape is probably because of distribution differences (uniform vs gaussian) in FlatTop or solver related (I'm still debugging that).

PR #353, opened 2026-04-09
Fix offset application in multirank for `FROMFILE` distribution

### LINKED ISSUE #352: `FromFile` doesn't apply `R0`/`P0` offsets correctly in multirank
Just a quick fix, where the indices over which the offsets are applied in `FromFile` multi-rank need to match the range reserved for this rank.

aliemen:
This pull request refactors how new particles are inserted into the Kokkos views in the `FromFile::generateParticles` function. The main change is to use Kokkos subviews to directly target the range of newly created particles, which simplifies the indexing logic and makes the code more robust and readable.


rammann (approved):
lgtm

PR #355, opened 2026-04-10
329 port variablerfcavity from old opal

### LINKED ISSUE #329: Port VariableRFCavity from old Opal
The VariableRFCavity element from old Opal uses the time dependency objects to generate time varying electric fields.

Jonathan-A-Thompson:
The variable RF cavity that uses time dependency objects to define the time varying electric field has been ported from old Opal.


src/AbsBeamline/VariableRFCavity.cpp
```diff
@@ -0,0 +1,195 @@
...
+    const double integralF = frequencyTD_m->getIntegral(t) * Units::MHz2Hz;
+    const double phi       = phaseTD_m->getValue(t);
+    return computeField(R, E, E0, integralF, phi, halfWidth_m, halfHeight_m);
+}
+
+bool VariableRFCavity::apply(
+        const Vector_t<double, 3>& R, const Vector_t<double, 3>& /*P*/, const double& t,
+        Vector_t<double, 3>& E, Vector_t<double, 3>& /*B*/) {
+    const double E0        = amplitudeTD_m->getValue(t);
+    const double integralF = frequencyTD_m->getIntegral(t) * Units::MHz2Hz;
```
chrisrogers1234, line 131: Just because we had units error in MultipoleT; can you check what should be the correct units? I did not expect to see a unit conversion

src/AbsBeamline/VariableRFCavity.cpp
```diff
@@ -0,0 +1,195 @@
...
+    return computeField(R, E, E0, integralF, phi, halfWidth_m, halfHeight_m);
+}
+
+bool VariableRFCavity::apply() {
+    const auto pc          = RefPartBunch_m->getParticleContainer();
+    const auto R           = pc->R.getView();
+    const auto E           = pc->E.getView();
+    const auto t           = RefPartBunch_m->getT();
+    const double E0        = amplitudeTD_m->getValue(t);
+    const double integralF = frequencyTD_m->getIntegral(t) * Units::MHz2Hz;
```
chrisrogers1234, line 142: See units comment

src/Elements/OpalVariableRFCavity.cpp
```diff
@@ -0,0 +1,90 @@
...
+// (at your option) any later version.
+//
+// You should have received a copy of the GNU General Public License
+// along with OPAL. If not, see <https://www.gnu.org/licenses/>.
+//
+#include "Elements/OpalVariableRFCavity.h"
+
+#include "AbsBeamline/VariableRFCavity.h"
+#include "Attributes/Attributes.h"
+
```
chrisrogers1234, line 22: I recall units should be Hz, Volts, metres. As commented elsewhere, check that my recollection is correct and everything is consistent

chrisrogers1234 (changes requested):

src/AbsBeamline/VariableRFCavity.cpp
Jonathan-A-Thompson: This conversion (MHz to Hz) is required.  The documentation requires the time dependency to generate MHz.  The implementation definitely needs Hz as it is calculating an angle as an input to the sine function.

src/AbsBeamline/VariableRFCavity.cpp
Jonathan-A-Thompson: Definitely required.

src/Elements/OpalVariableRFCavity.cpp
Jonathan-A-Thompson: The frequency units in the implementation are unambiguously Hz, so conversion from the UI MHz is required.  However, the units of the electric field is a little more obscure as this is passed back to the depths.  There is an existing implementation of ConstantEFieldCavity which does convert to V/m before returning the electric field, but there is no regression test so this could also be wrong.   The dump fields command assumes the values returned from apply are in MV/m, but it also assumes the B field is in kGauss (which it is not, we need to fix that) so that is no guide either.  The documentation does claim the electric field is in MV/m internally, do we believe that?  Looking at the Boris pusher it appears to assume E is in V/m.

Jonathan-A-Thompson: Keeping the conversion.

chrisrogers1234 (approved):

PR #364, opened 2026-04-16
Exclusive `PC`/`GAMMA`/`ENERGY` or `FROMFILE` options

### LINKED ISSUE #363: `PC` option silently ignored when using `FROMFILE` distribution
## Summary

When a `DISTRIBUTION` uses `TYPE=FROMFILE`, the `PC` (or `ENERGY`/`GAMMA`) attribute on the `BEAM` command is silently ignored. The particle momenta come entirely from the file, and the reference particle momentum is derived from the mean of the loaded particles — not from `BEAM.PC`.

In the following example, the value of `pc` is completely ignored, without warning.
```
Dist_FF: DISTRIBUTION, TYPE=FROMFILE,
    FNAME = "input-distribution-gauss.txt",
    NPARTDIST = n_particles;

ES1: EMISSIONSOURCE, DISTRIBUTION = Dist_FF;
mySources: EMISSIONSOURCELIST = (ES1);

BEAM1:  BEAM, PARTICLE = ELECTRON, pc = P0, NPART = n_particles,
        BFREQ = rf_freq, BCURRENT = beam_bunch_charge * rf_freq * 1e6, SOURCES = mySources, CHARGE = -1;
```

## Initialization

The reference particle momentum (`refPartP_m`) is initialized in `ParallelTracker::execute()` (`src/Algorithms/ParallelTracker.cpp:204-210`):

```cpp
if (pc->getTotalNum() > 0) {
    // Particles already loaded → use their mean momentum
    pc->getRefPartP() = beamlineToLab.rotateTo(pc->getMeanP());
} else {
    // No particles yet → fall back to BEAM's PC
    const PartData& pref = *pc->getReference();
    const double P0 = pref.getP() / pref.getM();  // beta*gamma
    pc->getRefPartP() = beamlineToLab.rotateTo(
        Vector_t<double, 3>(0.0, 0.0, P0));
}
```

For FROMFILE, particles are **always loaded before this runs**, so the `else` branch is dead code.

The call sequence in `TrackRun::execute()` guarantees particles exist before the tracker starts:

```mermaid
flowchart TD
    A["TrackRun::execute()"] --> B["setupDistributionsAndSamplers()
    <i>Particles loaded from file here</i>
    (TrackRun.cpp:377-384)"]
    B --> C["bunchUpdate()
    <i>Moments computed</i>
    (TrackRun.cpp:387)"]
    C --> D["new ParallelTracker(...)
    (TrackRun.cpp:413)"]
    D --> E["itsTracker_m->execute()
    (TrackRun.cpp:417)"]
    E --> F{"getTotalNum() > 0 ?"}
    F -->|"YES (FROMFILE)"| G["refPartP = getMeanP()
    <i>From actual particle data</i>"]
    F -->|"NO (empty container)"| H["refPartP = PartData P0
    <i>From BEAM PC</i>"]

    style G fill:#2d6a4f,color:#fff
    style H fill:#6c757d,color:#fff
```

## Solution
When using the `FROMFILE` option for the distribution, we disable the `pc` option.

rammann:
As described in #363, when using the `FROMFILE` option for the distribution, the `PC`/`GAMMA`/`ENERGY` options in the input file is quietly ignored. This could lead to confusion in the future. 

This PR removes this possibility. 
closes #363

<h2 data-heading="Files changed">Files changed</h2>

File | Change
-- | --
`src/Structure/Beam.h `| Added `hasExplicitEnergy()` declaration
`src/Structure/Beam.cpp` | Added `hasExplicitEnergy()`, removed energy check from `execute()`
`src/Track/TrackRun.cpp` | Added `FROMFILE` vs energy validation in `setupDistributionsAndSamplers()`

```mermaid
flowchart TD
    subgraph Before["Before (Beam::execute)"]
        direction TB
        B1["Beam::execute()"] --> B2{"PC or ENERGY\nor GAMMA set?"}
        B2 -->|No| B3["ERROR: energy not set"]
        B2 -->|Yes| B4["OK — no distribution\ncontext available"]
        
        style B3 fill:#dc3545,color:#fff
        style B4 fill:#6c757d,color:#fff
    end

    subgraph After["After (TrackRun::setupDistributionsAndSamplers)"]
        direction TB
        A1["setDistType() called\non Distribution"] --> A2{"Distribution type?"}
        A2 -->|FROMFILE| A3{"beam->hasExplicit\nEnergy()?"}
        A2 -->|GAUSS / other| A4{"beam->hasExplicit\nEnergy()?"}
        
        A3 -->|Yes| A5["ERROR: FROMFILE cannot\nbe combined with\nPC/ENERGY/GAMMA"]
        A3 -->|No| A6["OK — momenta\ncome from file"]
        
        A4 -->|Yes| A7["OK — energy sets\nmean pz"]
        A4 -->|No| A8["ERROR: energy\nnot set"]

        style A5 fill:#dc3545,color:#fff
        style A8 fill:#dc3545,color:#fff
        style A6 fill:#2d6a4f,color:#fff
        style A7 fill:#2d6a4f,color:#fff
    end
```

The regression tests need to be updated to match this new requirement: https://github.com/OPALX-project/regression-tests-x/pull/27/

## Result:
Trying to specify `PC` and `FROMFILE` at the same time gives:
```
Error> 
Error> *** User error detected by function "TrackRun::setupDistributionsAndSamplers()"
Error>     FROMFILE distribution "DIST1_FF" cannot be combined with PC/ENERGY/GAMMA on the BEAM. Remove the energy attribute from the BEAM command — particle momenta are read from the file.
```

### DIFF
--- a/src/Structure/Beam.cpp
+++ b/src/Structure/Beam.cpp
@@ -144,13 +144,6 @@ void Beam::execute() {
     }
 
     update();
-    // Check if energy explicitly has been set with the BEAM command
-    if (!itsAttr[GAMMA] && !(itsAttr[ENERGY]) && !(itsAttr[PC])) {
-        throw OpalException(
-            "Beam::execute()",
-            "The energy hasn't been set. "
-            "Set either \"GAMMA\", \"ENERGY\" or \"PC\".");
-    }
 
     if (!(itsAttr[PARTICLE]) && (!itsAttr[MASS] || !(itsAttr[CHARGE]))) {
         throw OpalException(
@@ -244,6 +237,10 @@ bool Beam::isPhoton() const {
 
 double Beam::getFrequency() const {
     return Attributes::getReal(itsAttr[BFREQ]);
+}
+
+bool Beam::hasExplicitEnergy() const {
+    return itsAttr[GAMMA] || itsAttr[ENERGY] || itsAttr[PC];
 }
 
 double Beam::getChargePerParticle() const {
--- a/src/Structure/Beam.h
+++ b/src/Structure/Beam.h
@@ -86,6 +86,9 @@ public:
     /// Throws if `SOURCES` is not set.
     std::string getEmissionSourceListName() const;
 
+    /// True if PC, ENERGY, or GAMMA was explicitly provided by the user.
+    bool hasExplicitEnergy() const;
+
     /// Update the BEAM data.
     virtual void update();
 
--- a/src/Track/TrackRun.cpp
+++ b/src/Track/TrackRun.cpp
@@ -554,6 +554,26 @@ void TrackRun::setupDistributionsAndSamplers(
         opalDist->setDist();
         opalDist->setAvrgPz(avrgpz);
 
+        // FROMFILE distributions carry absolute momenta — the BEAM's
+        // PC/ENERGY/GAMMA would be silently ignored, so forbid the combination.
+        if (opalDist->getType() == DistributionType::FROMFILE) {
+            if (beam->hasExplicitEnergy()) {
+                throw OpalException(
+                    "TrackRun::setupDistributionsAndSamplers()",
+                    "FROMFILE distribution \"" + src->getDistributionName()
+                    + "\" cannot be combined with PC/ENERGY/GAMMA on the BEAM. "
+                      "Remove the energy attribute from the BEAM command — "
+                      "particle momenta are read from the file.");
+            }
+        } else {
+            if (!beam->hasExplicitEnergy()) {
+                throw OpalException(
+                    "TrackRun::setupDistributionsAndSamplers()",
+                    "The energy hasn't been set. "
+                    "Set either \"GAMMA\", \"ENERGY\" or \"PC\" on the BEAM command.");
+            }
+        }
+
         distrs_m.push_back(distRaw);
 
         // Build a sampler instance for this emission source.


aliemen:
Is P0 or Gamma in the `BEAM` used anywhere else in the code? Or just for the initial distribution momentum/the reference particle? 

In that case I would even propose removing that all together and only maintaining the initial P0 in the `EMISSIONSOURCE` structure. We could even mirror the P0/energy/gamma things in the `EMISSIONSOURCE`? Your checks would probably fit well into the emission source.

rammann:
The only, very specific, case where it is used is when you start the simulation with en empty container and a `FLATTOP` distribution. In this case it will set the `refPartP` to the specified `PC`.

aliemen:
@rammann Nice, in that case I would propose getting rid of the pc/gamma/energy from the `BEAM` completely? Having these parameters globally in my opinion doesn't make sense when we have different distributions. So putting it in the emission source might be more fitting.

rammann:
All regression tests are passing on CPU and GPU when using the updated [regression tests branch](https://github.com/OPALX-project/regression-tests-x/pull/27/).

rammann:
> @rammann Nice, in that case I would propose getting rid of the pc/gamma/energy from the `BEAM` completely? Having these parameters globally in my opinion doesn't make sense when we have different distributions. So putting it in the emission source might be more fitting.

I would prefer not to change the inputfile's syntax too much. 
Conceptually, I would say the distribution is more a mathematical object, while the `BEAM` is a physics object. In this sense I think the energy/momentum fits better into the `BEAM`. 

I'd be happy to make a separate issue regarding the placement of `PC`/`ENERGY` in the inputfile. 

What do you think?

aliemen (approved):
OK. then let's merge this - I think it's a good idea to have these guardrails. However, I would then also make an issue to rethink how the whole energy/`P0`/gamma is put into the simulation.

Currently we have more or less two different values that somehow do the same (the one inside `EMISSIONSOURCE` and the global one). I would suggest unifying this inside the source, but in a different PR.

Issue #371, opened 2026-04-21
Support allocating empty particle containers with reserved capacity for the decay process


rammann:
## Summary

The upcoming decay process needs to allocate a `PartBunch` particle container that starts with **zero particles** but has:

1. **Reserved capacity** for particles created later during decay
2. **Correctly computed macrocharge and macromass**

Today this is not possible in OPALX because both the reserved allocation size and the macro quantities are derived from a single `BEAM::NPART` attribute, which is required to be strictly positive. We need to decouple "capacity / macro-quantity reference count" from "initial particle count".

PR #372, opened 2026-04-21
Rename NPART to NALLOC

rammann:
### Context

The `BEAM` object's `NPART` attribute is misleadingly named. In practice it does **not** set the number of simulated particles — that comes from the `DISTRIBUTION`'s `NPARTDIST`. Instead, `NPART` in BEAM is used only to:
1. Size the per-rank allocation buffer in `PartBunch::PartBunch`, passed via `totalParticlesPerBeam`.
2. Compute the macro-charge/macro-mass via `Beam::getChargePerParticle()` / `getMassPerParticle()`.

This PR is the first step in addressing issue #371. The `NPART` is renamed to `NALLOC` which is used to determine the memory that needs to be allocated. The creation of actual particles is left to the distribution classes and determined by the `NPARTDIST` attribute.

### Testing
With the [updated inputfiles](https://github.com/OPALX-project/regression-tests-x/pull/33) this is passing all regression and unit tests on CPU and GPU.

### DIFF
--- a/src/Algorithms/ParallelTracker.cpp
+++ b/src/Algorithms/ParallelTracker.cpp
@@ -712,7 +712,7 @@ void ParallelTracker::emitFromEmissionSources(double t, double dt) {
         // itsBunch_m->bunchUpdate();
 
         // Sanity guard: the total number of macroparticles in the bunch must
-        // never exceed the globally configured BEAM::NPART value. Overshooting
+        // never exceed the globally configured BEAM::NALLOC value. Overshooting
         // this limit would trigger internal reallocations in the particle
         // container and silently drop already-tracked particles/delete their data in the particle
         // attributes. This is only a check for the number of local particles.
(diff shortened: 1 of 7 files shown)


aaadelmann (commented):
NALLOC does not belong to the beam. I suggest to make NALLOC and option and if NALLOC is not set we set NALLOC to NPARTDIST. 

Also NPARTDIST is over specifying:  NPARTDIST is on a DISTRIBUTION hence NPART is sufficient.

rammann:
Do you mean a global option like the following?
```
OPTION, STATDUMPFREQ = 1;
```

aaadelmann:
YES

rammann:
I agree that `NALLOC` in the `BEAM` object is not ideal, since the `BEAM` object should (just) contain the physics. 

However, I would prefer to be able to allocate memory individually for each particle container. 
Since each `BEAM` initialises a particle container, this would be a practical place to specify the necessary memory.

aaadelmann:
ok, I see your point! Will aprove but to me this hints to a structural problem in the code i.e. where we do things ... can be addressed later!

aaadelmann (approved):

rammann:
Thanks, should I make an issue regarding this?

aaadelmann:
yes

PR #376, opened 2026-04-21
Migration of `TravelingWave` RF cavity #358

### LINKED ISSUE #358: Add Travelling Wave RF Cavity Element to OPALX
OPALX currently lacks RF cavity elements required to fully model SwissFEL for the Dense Coulomb Clouds project. The standing wave `RFCavity` has recently been introduced in [#302](https://github.com/OPALX-project/OPALX/pull/302), providing a GPU-compatible foundation based on the legacy OPAL implementation.

The next missing accelerator element needed to complete the SwissFEL lattice is the **travelling** wave `RFCavity`.

The existing travelling wave cavity model in OPAL serves as a reference implementation but is CPU-only and tightly coupled to legacy code. A modernized, GPU-compatible version is required for OPALX.

In addition, this work will possibly require migrating and adapting the `AstraDynamic` field map, as required for existing SwissFEL field maps.

**Goals of this issue:**
- Design and implement a **travelling** wave `RFCavity` element in OPALX
- Ensure compatibility with GPU execution and OPALX architecture
- Migrate and integrate the `AstraDynamic` field map, as required for existing SwissFEL field maps
- Implement unit tests for the **travelling** wave `RFCavity` and `AstraDynamic` field map
- Add a regression test in the appropriate repository

MichZampetakis:
This PR adds support for the `TravelingWave` RF cavity in OPALX and migrates the `Astra1DDynamic` fieldmap implementation so it can be used in the Kokkos/GPU path.

The main goal was to reproduce the OPAL `TravelingWave` behavior in OPALX while keeping the field evaluation device-safe and compatible with GPU execution.

During validation against OPAL, a major discrepancy was observed: the cavity produced much less energy gain than expected. The issue was traced to the `Astra1DDynamic` **Fourier reconstruction**, not to the `TravelingWave` region logic itself. This PR includes a way to bypass this.

### TravelingWave in OPALX

* Added `TravelingWave` element to OPALX
* Added device-safe field application using `Astra1DDynamic::applyTravelingWave(...)`
* Preserved OPAL-style entry / core / exit region logic
* Preserved OPAL-style phase handling for:
  * entry field
  * two core contributions
  * exit field

### Astra1DDynamic migration

* Migrated `Astra1DDynamic` to OPALX
* Replaced raw pointer coefficient storage with `Kokkos::DualView<double*>`
* Added host/device-compatible `computeField(...)`
* Added host/device-compatible `computeTravelingWaveField(...)`
* Added `applyField(...)` and `applyTravelingWave(...)` Kokkos paths

---

## Root cause of the validation discrepancy

The reduced energy gain was initially suspected to come from:

* TravelingWave region boundaries
* reference-particle timing / phase usage
* fieldmap normalization
* host vs device application differences

These were checked and ruled out.

The actual issue was in the **Fourier coefficient reconstruction of** `Astra1DDynamic`.

### Problem

The original OPALX migration used Fourier coefficients that produced an incorrect reconstructed on-axis field:

* wrong sign in parts of the waveform
* wrong endpoint behavior
* wrong amplitude/shape relative to the raw fieldmap

This was confirmed by comparing:

* raw fieldmap samples
* reconstructed `Ez(z)` from `computeField(...)`

### Fix

The fieldmap preprocessing was changed to compute Fourier coefficients **explicitly** from the mirrored periodic sample array instead of relying on the packed GSL real-FFT output interpretation.

The critical correction was the sine-coefficient convention:

* reconstruction uses
  `Ez(z) = a0 + sum_l [ a_l cos(l*kz) - b_l sin(l*kz) ]`
* therefore the stored sine coefficient must match that convention

After this fix, the reconstructed field matches the raw on-axis field profile closely.

---

### `Astra1DDynamic::readMap()`

* reads and filters strictly increasing raw `z` samples
* stores raw on-axis samples
* interpolates to an equidistant grid
* constructs a mirrored periodic array
* computes Fourier coefficients explicitly
* stores coefficients in `Kokkos::DualView`

### Notes

* The previous GSL real-FFT path is left commented as a future optimization candidate
* `getOnaxisEz()` remains file-based / normalized for compatibility with existing TravelingWave autophase logic

---

### Unit tests

* Added unit tests for `Astra1DDynamic`, covering parsing, field evaluation, field accumulation, derivatives, and on-axis field behavior (included in this PR)
* Added unit tests for `TravelingWave`-related field application paths, including entry, core, exit, and out-of-range behavior (included in this PR)

### Regression test

* Performed a regression test using a SwissFEL TravelingWave cavity case
* Compared OPALX against OPAL and verified that the corrected implementation reproduces the expected cavity behavior
_Files attached here, PR in [the regression test repo](https://github.com/OPALX-project/regression-tests-x) will be opened._
[TWCavity.zip](https://github.com/user-attachments/files/26943650/TWCavity.zip)

<p align="center">
  
  
</p>

Using the exact same particle distribution as in OPAL does not noticeably improve the agreement. A remaining source of discrepancy may be the Fourier preprocessing, since the current OPALX implementation does not use the original GSL FFT-based path and instead computes the coefficients explicitly.

---

## Important behavior note

For TravelingWave, the effective tracked element length is determined from:

* fieldmap longitudinal span
* `NUMCELLS`
* `MODE`
* entry/exit field regions

So the runtime effective length may differ from the nominal input `LENGTH` parameter. This is consistent with the current `TravelingWave` initialization logic. In this example, `L=2.80` but the tracked cavity length appears to be ~4.1 m


src/Fields/Astra1DDynamic.cpp
```diff
@@ -0,0 +1,668 @@
...
+            } 
+            // else {
+            //     ++skippedValues;
+            // }
+        }
+
+        num_gridpz_m = acceptedValues;
+        lines_read_m = 0;
+
+        // parsing_passed =
```
aaadelmann: In this file there is a lot of commented out code. Please remove of indicate why we should kepp it.

src/Fields/Astra1DDynamic_fast.h
```diff
@@ -0,0 +1,59 @@
...
+//
+// OPAL is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// You should have received a copy of the GNU General Public License
+// along with OPAL. If not, see <https://www.gnu.org/licenses/>.
+//
+#ifndef CLASSIC_AstraFIELDMAP1DDYNAMICFAST_HH
```
aaadelmann: Remove all references to classic CLASSIC

src/AbsBeamline/RFCavity.cpp
```diff
@@ -1,8 +1,8 @@
 /**
```
aaadelmann, line 1: Are CPU and GPU tests done?

aaadelmann (changes requested):

src/Fields/Astra1DDynamic.cpp
MichZampetakis: I removed the old comments from`Astra1DDynamic::Astra1DDynamic():Fieldmap()`. I will push soon.

In `Astra1DDynamic::readMap()`, the remaining commented lines are:
```
// Future optimization candidate:
// gsl_fft_real_wavetable* real =
//     gsl_fft_real_wavetable_alloc(2 * num_gridpz_m);
// gsl_fft_real_workspace* work =
//     gsl_fft_real_workspace_alloc(2 * num_gridpz_m);
```
I kept these intentionally because they document the original GSL FFT-based path mentioned in the PR. The current implementation uses explicit coefficient construction, but the commented lines indicate a possible future optimization if we want to revisit the FFT-based preprocessing.

In `Astra1DDynamic::applyTravelingWave()`, the commented `Kokkos::parallel_for()` was my first implementation which I changed to be able to call some unit tests without initializing particles. If you prefer the uncommented version, I can remove the commented one.

src/Fields/Astra1DDynamic_fast.h
MichZampetakis: At the moment, only `Astra1DDynamic` was migrated, since that was needed to support the `TravelingWave` element in OPALX. `Astra1DDynamic_fast` has not been migrated yet, but I can do it next.

src/Fields/Astra1DDynamic.cpp
aaadelmann: Yes keep the "Future optimization candidate:" but delete the others please

aaadelmann:
The fast version can be done later. Lets get the TW out with regression tests running on CPU/GPU

src/Fields/Astra1DDynamic.cpp
MichZampetakis: OK, done!

src/AbsBeamline/RFCavity.cpp
MichZampetakis: I did a regression test on CPU (Serial/Parallel). @aliemen just checked the unit tests on CPU Serial and CUDA and they look good. I will clean-up my regression test and open a PR soon.

aaadelmann (approved):

src/Fields/Astra1DDynamic.cpp
```diff
@@ -0,0 +1,527 @@
...
+
+    auto FourCoefs_device = FourCoefs_m.view_device();
+
+    auto Rview = pc->R.getView();
+    auto Eview = pc->E.getView();
+    auto Bview = pc->B.getView();
+
+    Kokkos::parallel_for(
+        "Astra1DDynamic::applyTravelingWave",
+        ippl::getRangePolicy(Rview),
```
aliemen: Please use the particle count directly (which you get from the `pc`) instead of `ippl::getRangePolicy(Rview)`

src/Fields/Astra1DDynamic.cpp
```diff
@@ -0,0 +1,527 @@
...
+        KOKKOS_LAMBDA(const int i)
```
aliemen: `size_t` instead of `int`

src/Fields/Astra1DDynamic.h
```diff
@@ -0,0 +1,334 @@
...
+     * 'this' pointer not allowed on device)
+     */
+    bool isInside(const Vector_t<double, 3> &r) const override {
+        return r(2) >= zbegin_m && r(2) < zend_m;
+        // && sqrt(r(0) * r(0) + r(1) * r(1)) < rend_m;
+    }
+
+    template <class ViewType>
+    KOKKOS_INLINE_FUNCTION
+    static void computeField(
```
aliemen, line 97: Can you put it in the `.cpp` instead of the header?

src/Fields/Astra1DDynamic.h
```diff
@@ -0,0 +1,334 @@
...
+        E(1) += EfieldR * R(1);
+        E(2) += ez + 4.0 * f * RR2;
+
+        B(0) -= BfieldT * R(1);
+        B(1) += BfieldT * R(0);
+    }
+
+    template <class ViewType>
+    KOKKOS_INLINE_FUNCTION
+    static void computeTravelingWaveField(
```
aliemen, line 148: Same here with the `.cpp`

src/Fields/FM2DDynamic.cpp
```diff
@@ -209,22 +206,21 @@ void FM2DDynamic::readMap() {
         FieldstrengthBt_m.sync<Kokkos::DefaultExecutionSpace>();
 
         *ippl::Info << level3 
-            << typeset_msg("read in fieldmap '" + Filename_m + "'", "info")
-            << endl;
+                    << typeset_msg("read in fieldmap '" + Filename_m + "'", "info")
+                    << endl;
     }
```
aliemen: Please use `Inform`. E.g.:
```cpp
Inform m("FM2DDynamic::readMap");
m << level3 << "Read in fieldmap" ...
```

src/Fields/FM2DDynamic.cpp
```diff
@@ -209,22 +206,21 @@ void FM2DDynamic::readMap() {
...
 
         FieldstrengthEz_m = Kokkos::DualView<double*>();
         FieldstrengthEr_m = Kokkos::DualView<double*>();
         FieldstrengthBt_m = Kokkos::DualView<double*>();
 
         *ippl::Info << level3 
-            << typeset_msg("freed fieldmap '" + Filename_m + "'", "info") 
-            << endl;
+                    << typeset_msg("freed fieldmap '" + Filename_m + "'", "info") 
+                    << endl;
```
aliemen: Same here with `Inform`

src/AbsBeamline/RFCavity.cpp
```diff
@@ -12,7 +12,7 @@
 #include <filesystem>
 #include "AbsBeamline/BeamlineVisitor.h"
 #include "Fields/Fieldmap.h"
-#include "Fields/FM2DDynamic.h"
+#include "Fields/FM2DDynamic.h" // is it needed?
```
aliemen: I don't know. But if not, feel free to remove the line.

src/AbsBeamline/TravelingWave.cpp
```diff
@@ -1,30 +1,22 @@
...
+#include "PartBunch/PartBunch.h"
 #include "Physics/Units.h"
-#include "Utilities/GSLSpline.h"
+#include "Utilities/GeneralClassicException.h"
+// #include "Utilities/GSLSpline.h"
 
-#include <fstream>
-#include <iostream>
+// #include <fstream>
+// #include <iostream>
```
aliemen: Can be removed if not needed

src/Fields/Astra1DDynamic_fast.h
```diff
@@ -0,0 +1,59 @@
+//
```
aliemen: If it's not migrated, you can just remove the whole file again (if I see it correctly, then it's also not present in the current master). Same with the corresponding `.cpp`. Perhaps make an issue that it's not migrated.

aliemen (changes requested):
A few small suggestions. After that, we can wait for the regression test

src/Fields/Astra1DDynamic.cpp
MichZampetakis: Modified as advised in `Astra1DDynamic` and `FM2DDynamic` fieldmaps:
```
const size_t nLocal = pc->getLocalNum();

Kokkos::parallel_for(
     "Astra1DDynamic::applyTravelingWave", nLocal,
      KOKKOS_LAMBDA(const size_t i)
      {
      }
```

src/Fields/Astra1DDynamic.cpp
MichZampetakis: Resolved together with the comment above.

src/Fields/FM2DDynamic.cpp
MichZampetakis: Replaced `*ippl::Info` messages with `Inform` in `RFCavity`, `TravelingWave`, `FM2DDynamic` and `Astra1DDynamic`.

src/Fields/FM2DDynamic.cpp
MichZampetakis: Included in the previous comment.

src/AbsBeamline/RFCavity.cpp
MichZampetakis: It is needed in the end.

src/Fields/Astra1DDynamic_fast.h
MichZampetakis: `Astra1DDynamic_fast` files were deleted and an issue was opened to follow-up the fieldmap migration.

src/Fields/Astra1DDynamic.h
MichZampetakis: I think both `computeField()` and `computeTravelingWaveField()` need to stay in the header. When they were implemented in the `.cpp` (for my initial implementation of the `RFCavity`), there were device compiler errors on the GPU path. @aaadelmann worked around this by making them templated `KOKKOS_INLINE_FUNCTION` helpers in the header so the device compiler can see their definitions.

If you know a clean way to move them back into the .cpp without breaking the Kokkos/GPU build, I’d be happy to discuss it.

src/Fields/Astra1DDynamic.h
MichZampetakis: I resolve this, as it is included in the comment above.

src/Fields/Astra1DDynamic.h
aaadelmann: Why not keeping this for the moment and make an issue. I agree with @aliemen it would be nicer in the .cpp

aliemen (approved):
Nice, looks good now! All regression tests including unit tests pass on CPU/GPU. Also quickly ran your new regression test multirank - all good.

PR #379, opened 2026-04-21
fix warning on missing overwrite

aaadelmann:
fix warning on missing overwrite


src/Algorithms/DefaultVisitor.h
```diff
@@ -67,7 +67,7 @@ class DefaultVisitor : public BeamlineVisitor {
     void visitMultipole(const Multipole&) override;
 
     /// Apply the algorithm to to an arbitrary multipole.
```
mohsensadr: please also fix the comment: repeated "to to"

mohsensadr (changes requested):

src/Algorithms/DefaultVisitor.h
aaadelmann: done

mohsensadr (approved):

PR #389, opened 2026-04-24
Image charges, Shifted Green's Solver, `mirrorField` helper, `DirichletPlaneWriter`, Emission Fixes

### LINKED ISSUE #319: Implement image charges/Dirichlet BCs for full AWAGun
For a full `AWAGun` run in OPALX, there are still image charges or (later) Dirichlet BCs missing at the gun source, otherwise results will be [a bit off](https://github.com/OPALX-project/OPALX/pull/308#issuecomment-4129758930). 

In my opinion, the best approach would be the following: create an interface e.g. for `PartBunch` that looks something like this:
```cpp
void applyGridModificationElement(... elem, newMin, newMax) {
    elem.transformAttr(pc);
    grid.resize(newMin, newMax);
    scatter(...); // now scatters the transformed bunch
    elem.backTransformAttr(pc);
}
```
This would be applied for example after every scatter step. The grid modification element would in this case then mirror bunch positions and change the charge. The scatter would automatically "apply" the image charges. Then, the element handles reseting the bunch to its original state. That way we don't have and additional memory footprint. Using the `ParticleContainer` ensures that you can link different particle containers to this image element. Then we could create a new element, e.g. called `IMAGECHARGE` that handles the bunch transformations for that. 

Perhaps we could also have a new structure (or array of such `IMAGEELEMENT`s) that is passed to the FieldSolver (since it ultimately is a modification done to the `solve`). 

The upside of this approach would be that it could also "mirror" a bunch in the interactions as required by #300.

aliemen:
closes #319 

This PR implements two ways of activating Dirichlet BCs at a cathode plane and includes various bug fixes detailed below. The most important part is the shifted Green's function implementation that allows the binned field solver to be used in the same way as OPAL's `GunSource` together with FlatTop. 

The big test which I am trying to reproduce here is `AWAGun-1.in`. Previous inaccuracies outlined in #319 are mostly due to differing initial conditions. Here I'll take OPAL as a reference solution that I am trying to replicate. Since I only want to focus on the shifted Green's (Dirichlet BC with image charge ) solver in this PR, I did the following to get matching initial conditions:
- Verify that the open solver is verifiable by a test with analytically solution, see [here](https://github.com/IPPL-framework/ippl/pull/494#issue-4302940805).
- Use an injected `FROMFILE` distribution to get the same initial phase space.
- Verify that the no space charge leads to the same results, see [here](https://github.com/OPALX-project/OPALX/issues/319#issuecomment-4222630430).
- Hack the shifted Green's solver path into OPAL for the non-FlatTop distribution.  
- Use only 1 bin: relativistic transformation before solve, but approximate with non-relativistic energy spread.

The easiest comparison is to run [`OPALX-1E4-1bin-SG`](https://github.com/user-attachments/files/27053368/OPALX-1E5-1bin-SG.txt) and [`OPAL-1E4-1bin-SG`](https://github.com/user-attachments/files/27053370/OPAL-1E4-1bin-SG.txt) and compare (note that OPAL here needs to have the forcefully activated shifted Green's path in the code). With `1E5` particles and `N=32` grid points, we get: 

awa-1bin-shiftedgreens-1E5-32

This shows that (with enough particles and grid points) OPAL and OPALX agree well - showing that the Dirichlet BCs are correctly implemented. With only `1E4` particles and `N=16` the results is a bit worse:

awa-1bin-shiftedgreens

Note that the second plot was done with OPALX singlerank. The first plot (to save time) was done on 8 ranks with OPALX - showing that the critical part, the [`mirrorField` helper](https://github.com/OPALX-project/OPALX/blob/7e778100d0f096a904f2aad8a1a763b297fc920e/src/PartBunch/FieldMirror.hpp#L45), works perfectly fine multirank. Apart from that, I want to compare the same to a 2 rank GH200 run:

multirank_gh200

This should show that the same path also works nicely multirank on GPU. Now regarding the input file, I added some flags to the `EmissionSource` that define how the Dirichlet BCs are used. Now the following options are available:
```
ES1: EMISSIONSOURCE, DISTRIBUTION = Dist,
    ZEROFACE_R0Z = false,
    ZEROFACE_MAXSTEPS = 19000,
    SHIFTED_GREENS_FUNCTION = true,
    R0Z=R;
    //ZEROFACEPLANEDUMP = 100
```
Here a list of what these do:
- `ZEROFACE_R0Z`: Activates explicit image charge computation. As long as the Dirichlet BCs are active, it doubles the grid, inverts the charges, mirrors the distribution, does a second scatter/solve/gather routine. Default `false`.
- `ZEROFACE_MAXSTEPS`: Defines how many timesteps the Dirichlet BCs are active. Perhaps we can later replace this with some form of automated switch. Default `0`.
- `SHIFTED_GREENS_FUNCTION`: Activates the shifted Greens solver path. Can only be true of `ZEROFACE_R0Z=false`. Default `false`.
- `R0Z`: Same as before, but also defines at which z position the cathode plane is located. _If necessary, this could later become its own flag (something like `CATHODE_R0Z`)._ Default `0`.
- `ZEROFACEPLANEDUMP`: Timestep frequency at which dirichlet plane diagnostics are printed (only for the explicit image charges, since the shifted Greens function never actually calculates the potential at the cathode). Writes a 2D ASCII field to a file containing the potential values at the cathode plane. Also prints out its mean and rms potential at the cathode plane. Works single rank on CPU and GPU, otherwise it's deactivated. Uses the same format as the write function in `FieldSolver.cpp`.  
 
That concludes the results of this PR. Next I want to go more into the details of what was implemented/changed.

---

Now to the (quite long, probably not that interesting, but should all at least be mentioned) individual changes.

### 1. Emission Process Fixes

**Coordinate frame transformation at emission** (`ParallelTracker::emitFromEmissionSources`):
Particles are now transformed into the gun-local coordinate system before the sampler runs, then back to the reference frame afterwards:
```cpp
pc->transformBunch(refToGun);
...sampler->sample(...)...
pc->transformBunch(refToGun.inverted());
```
Previously particles were generated directly in the reference frame, which gave wrong positions whenever the gun is not at the origin or is rotated.

**Half-push z-correction at birth** (`FlatTop::generateUniformDisk`):
New particles are born between `timeIntegration1` and `timeIntegration2`, so they only receive the second half-push. Without correction the intra-batch z-spread is half the inter-batch spacing, producing visible disc artefacts. The fix applies the missing half-push at birth:

$$
z_\text{init} \mathrel{+}= \tfrac{1}{2} \beta_z(p_\text{birth}) c f\cdot\mathrm{d}t
$$

where $f \sim U(0,1)$ is the fractional timestep already stored in `dtView(j)`.

**P0 range guard** (`SamplingBase`): The reference momentum `P0` is now only applied to newly created particles and not reset every timestep, preventing momentum resets during continuous emission.

---

### 2. Explicit Image Charges with Dirichlet Plane Diagnostics

A new `ImageChargeScatterController` class centralises mirror-particle logic (previously scattered across `PartBunch`). It exposes three scatter modes used by `prepareRhoForBin`:

| Mode | Scatter target |
|---|---|
| `PrimaryAndImage` | Both real and mirror charges in one pass (legacy) |
| `PrimaryOnly` | Real charges only |
| `ImageOnly` | Mirror charges only ($-q$, position reflected across $z_p$) |

The binned field solve now runs **two separate passes** per bin when image charges are active:
- Pass 1 (`PrimaryOnly`, `bFieldSign = +1`): standard self-field.
- Pass 2 (`ImageOnly`, `bFieldSign = -1`): image field. The $-1$ sign accounts for the mirror bunch moving with velocity $-\mathbf{v}$, so $\mathbf{B}_\text{image} \propto -\mathbf{v} \times \mathbf{E}^{\prime}_\text{image}$.

**Dirichlet plane diagnostics** (`DirichletPlaneWriter`): A new optional writer dumps the interpolated potential $\phi(z = z_p)$ at the cathode plane every `ZEROFACEPLANEDUMP` steps, reporting mean and variance so users can verify $\phi \approx 0$ is being enforced.

**New EMISSIONSOURCE keywords** (already summarized above):
- `ZEROFACE_R0Z`: enable explicit image charges at the cathode plane $z = R_{0z}$.
- `SHIFTED_GREENS_FUNCTION`: enable the shifted-GF Dirichlet correction (see below).
- `ZEROFACE_MAXSTEPS`: step budget; image charges deactivate automatically after this many steps (0 = unlimited).
- `ZEROFACEPLANEDUMP`: potential dump frequency at the Dirichlet plane.

Both `ZEROFACE_R0Z` and `SHIFTED_GREENS_FUNCTION` are mutually exclusive and raise an `OpalException` if both are set to `true`.

---

### 3. Tracker Loop Reordering

Emission is moved **before** external-field evaluation to match legacy OPAL's `ParallelTTracker` step ordering:

**Before:**
```
external fields → setTime → emit → incrementT
```

**After:**
```
selectDT → setTime → emit → external fields → incrementT
```

This ensures newly emitted particles experience external fields in their creation step. The `globalEOL_m` flag is also reset at the start of each step (transient out-of-bounds from invalid mesh bounds right after first emission must not persist) and the MPI `reduce` → `allreduce` fix ensures all ranks receive the end-of-line result.

---

### 4. Timestep Size Assignment Fixes

- **`Gaussian::emitParticles`**: Newly created particles were Kokkos-zero-initialised, giving `dt = 0`. `switchToUnitlessPositions(use_dt_per_particle=true)` divides by `dt`, producing NaN positions. Fix: assign `dt = tEnd - t0` (remaining fraction of the current step) immediately after `generateParticles`.
- **`FlatTop` / `t0 != 0` injection**: `switchToUnitless` previously divided by `dT` before `dt` was assigned, causing `0/0 = NaN` for the first injection batch. Fixed by ensuring `dt` is set before any unitless-position conversions.
- **BCHARGE safeguard**: If the field solver is active but `chargePerParticle == 0`, an early `OpalException` is thrown with a clear message pointing to the missing `BCHARGE` keyword, instead of silently propagating NaN through `rho → E → particles`.

---

#### Motivation

The explicit image-charge path requires the computational mesh to physically straddle the cathode plane, which introduces huge inaccuracies later on. The shifted Green's function encodes the image contribution analytically in the convolution kernel, so it remains correct at any bunch-cathode distance.

#### Physics derivation

Standard free-space Poisson in the bunch rest frame:

$$
\nabla^2 \phi'(\mathbf{r}') = -\frac{\rho'(\mathbf{r}')}{\varepsilon_0}, \qquad \mathbf{E}' = -\nabla'\phi'
$$

For a grounded cathode at $z = z_p$ the image potential satisfies:

$$\phi_\text{image}(\mathbf{r}) = -\phi_\text{real}(R_p(\mathbf{r})), \qquad R_p(x,y,z) = (x,y, 2z_p - z)$$

Convolving $\rho_\text{real}$ with a **shifted** free-space kernel $G(\mathbf{r} - \mathbf{s} - \mathbf{r}'')$ produces a potential displaced by $\mathbf{s}$. Choosing $\mathbf{s} = (0,0,s_z)$ such that the grid-flip $R_g$ maps $R_g(\mathbf{r}) - \mathbf{s} = R_p(\mathbf{r})$ gives:

$$s_z = 2\left(z_c^\text{rest} - z_p\right), \qquad z_c^\text{rest} = z_\text{origin} + \tfrac{N_z}{2} h_z^\text{orig} \gamma_\text{bin}$$

#### Vector-field sign rule

The solver outputs $\mathbf{E}' = -\nabla'\phi'$ directly. Flipping the index along $z$ and differentiating through $R_g$ introduces a component-dependent sign:

$$
\mathbf{E}_{\text{image}}(\mathbf{r}) = \Sigma \mathbf{E}_{\text{shifted}}(R_g(\mathbf{r})), \quad \Sigma = \mathrm{diag}(-1,-1,+1)
$$

#### Program flow

```mermaid
flowchart TD
    Entry["computeSelfFields(bunch)"]
    Entry --> HasBins{"has binning?"}
    HasBins -- no --> Legacy["computeLegacySelfFields (no correction)"]
    HasBins -- yes --> BinLoop["for each merged bin b"]

    BinLoop --> Gamma["computeGammaBinGlobal → {pmean, γ_b}"]
    Gamma --> Mode{"correction mode?"}

    Mode -- "none" --> P1
    Mode -- "ZEROFACE_R0Z" --> IP1
    Mode -- "SHIFTED_GREENS" --> SP1

    subgraph Primary["Pass 1: primary self-field (all modes)"]
        P1["prepareRhoForBin(PrimaryOnly)"]
        P1 --> P2["mesh.setMeshSpacing(hr_z × γ_b)"]
        P2 --> P3["E = 0; runSolver()"]
        P3 --> P4["accumulateFieldToTemp(bFieldSign=+1, flipAxis=-1)"]
        P4 --> P5["mesh.setMeshSpacing(hr_z_orig)"]
    end

    subgraph ImagePath["Pass 2A: explicit image charges"]
        IP1["prepareRhoForBin(ImageOnly)\n(mirror positions, flip charge sign)"]
        IP1 --> IP2["E = 0; runSolver()"]
        IP2 --> IP3["accumulateFieldToTemp(bFieldSign=-1, flipAxis=-1)"]
    end

    subgraph ShiftedPath["Pass 2B: shifted Green's function"]
        SP1["prepareRhoForBin(PrimaryOnly)\n(re-scatter primary)"]
        SP1 --> SP2["*(rho) = -(*(rho))\n(negate: image has opposite polarity)"]
        SP2 --> SP3["compute shift: s_z = 2(z_c^rest - z_p)"]
        SP3 --> SP4["runShiftedOpenSolver(shift):\n  shiftedGreensFunction(s)\n  solve()\n  greensFunction() (restore kernel)"]
        SP4 --> SP5["buildFlippedZSlab(E')\n(MPI peer-rank z-flip exchange)"]
        SP5 --> SP6["accumulateFieldToTemp(bFieldSign=-1, flipAxis=Dim-1)\n(apply σ_i sign rule, then Lorentz transform)"]
    end

    P5 --> EndBin
    IP3 --> EndBin
    SP6 --> EndBin

    EndBin["end bin loop"]
    EndBin --> Gather["gatherFromTempToParticles\n(pc→E from Etmp, pc→B from Btmp)"]
```

#### Full equation (binned + shifted Green's function)

$$
\mathbf{E}_\text{lab} = \sum_b \Lambda_{\gamma_b,\mathbf{v}_b} \left[\frac{1}{\varepsilon_0\gamma_b}\int G(\mathbf{r}-\mathbf{r}') \rho^{(b)}_+(\mathbf{r}') \mathrm{d}^3r' + \mathbf{E}^{(b)}_\text{image}(\mathbf{r})\right]
$$

where

$$
E^{(b)}_{\text{image},i}(\mathbf{r}) = \sigma_i \cdot E^{(b)}_{\text{shifted},i}(R_g(\mathbf{r})), \qquad
\mathbf{E}^{(b)}_\text{shifted}(\mathbf{r}) = \frac{1}{\varepsilon_0\gamma_b}\int G(\mathbf{r}-\mathbf{s}_b-\mathbf{r}') \rho^{(b)}_+(\mathbf{r}') \mathrm{d}^3r'
$$

and $\Lambda_{\gamma,\mathbf{v}}$ is the quasi-static Lorentz transform:

$$
\mathbf{E}_\text{lab} = \gamma\mathbf{E}' - (\gamma-1)(\mathbf{E}'\cdot\hat{\mathbf{w}})\hat{\mathbf{w}}, \qquad \mathbf{B}_\text{lab} = \frac{\gamma}{c^2} \mathbf{v}\times\mathbf{E}'
$$

#### Multi-rank z-flip (`buildFlippedZSlab`)

Under `PARFFTZ=true` the z-axis is distributed across MPI ranks. The global flip $k \to N_z^\text{global} - 1 - k$ generally crosses ranks. `buildFlippedZSlab` performs a pairwise-exchange pass: each rank computes which z-slabs it needs from each peer, packs them into byte buffers, and exchanges via `MPI_Isend`/`MPI_Irecv` (tag `0x5F1F`). The device view is host-staged (`create_mirror_view_and_copy` in, `deep_copy` out) so the exchange works identically on SERIAL, OpenMP, CUDA, and HIP builds without requiring device-aware MPI. This is a performance concern and might be adressed later on. 

---

### Files Changed (non-trivial)

| File | Change |
|---|---|
| `PartBunch/BinnedFieldSolver.{h,tpp}` | Two-pass solve, shifted-GF path, Dirichlet diagnostics |
| `PartBunch/ImageChargeScatterController.{h,tpp}` | New class: mirror-scatter with PrimaryOnly/ImageOnly/PrimaryAndImage modes |
| `PartBunch/FieldMirror.hpp` | New generic `mirrorField` helper used by `buildFlippedZSlab` |
| `PartBunch/FieldSolver.{hpp,cpp}` | `runShiftedOpenSolver(shift)`: shifted kernel → solve → restore |
| `Structure/EmissionSource.{h,cpp}` | New keywords: `ZEROFACE_R0Z`, `SHIFTED_GREENS_FUNCTION`, `ZEROFACE_MAXSTEPS`, `ZEROFACEPLANEDUMP` |
| `Structure/DirichletPlaneWriter.{h,cpp}` | New: interpolated potential dump at cathode plane |
| `Track/TrackRun.cpp` | `configureImageChargeFromSources`, image/shifted-GF wiring, mutual-exclusion guards |
| `Algorithms/ParallelTracker.cpp` | Loop reordering (emit before external fields), EOL flag reset, `allreduce` fix |
| `Distribution/FlatTop.cpp` | z half-push correction at birth |
| `Distribution/Gaussian.cpp` | `dt` assignment for newly created particles |
| `PartBunch/PartBunch.{h,cpp}` | BCHARGE safeguard, mesh z-resolution doubling while image charges active |


aliemen:
This PR is ready for review. ~For your information: it requires IPPL-framework/IPPL#495 to be merged first. If you want to test it locally, compile with `-DIPPL_GIT_TAG=mirror-field-implementation`.~

Another fyi: with this branch, the previously disabled [`Drift-4-multi-emit-open`](https://github.com/OPALX-project/regression-tests-x/pull/26) works again. So once this PR goes through, I'll redo FlatTop reference (because of the above sampling reference frame fixes) and re-activate `...-multi-emit-open`. 

Edit: `mirror-field-implementation` is merged - ~rerunning the status checks now~ everything works.

aliemen:
All unit tests and regression tests pass on CPU and GPU (GH200) - except for FlatTop, but that's intended due to the emission fix mentioned above.

aaadelmann (commented):
```
ES1: EMISSIONSOURCE, DISTRIBUTION = Dist,
    ZEROFACE_R0Z = false,
    ZEROFACE_MAXSTEPS = 19000,
    SHIFTED_GREENS_FUNCTION = true,
    R0Z=R;
```

_SHIFTED_GREENS_FUNCTION_ is a solver attribute I think this is the wrong place.

I would also argue the other do better belong to the fieldsolver 
_ZEROFACE_R0Z = false, ZEROFACE_MAXSTEPS = 19000, SHIFTED_GREENS_FUNCTION = true_

aliemen:
@aaadelmann semantically - yes I agree. I put it in the EmissionSource, because otherwise it might not be clearly defined at which `R0Z` the cathode is located (there might be multiple distributions and now it takes `R0Z` from the EmissionSource placement). 

If I put it in the solver, I would need an additional `ZEROFACE_R0Z` attribute in the solver. Perhaps my suggestion would even be to add another `ZEROFACE` structure that's passed to the FieldSolver. E.g. something like this:
```
ZF: ZEROFACE,
    R0Z=R, // defaults to 0
    ZEROFACE_R0Z = false, // defaults to false 
    ZEROFACE_MAXSTEPS = 19000, // defaults to MAXSTEPS
    SHIFTED_GREENS_FUNCTION = true; // defaults to true

// then pass it to the solver:
FS1: Fieldsolver, TYPE=OPEN, BINS=BINS1,
    ZEROFACE=ZF, 
    NX=N, NY=N, NZ=N,  
    ...
```
I think this is the most versatile solution and more or less mirrors the "`GUNSOURCE`" element semantics from old OPAL. Should I implement it that way _(in that case I'll probably also change the attribute names slightly)_?

aaadelmann:
Lets keep this for the moment maybe just open an issue.

src/PartBunch/PartBunch.cpp
```diff
@@ -12,35 +13,33 @@
...
+    : ippl::PicManager<
+              T, Dim, ParticleContainer<T, Dim>, FieldContainer<T, Dim>, LoadBalancer<T, Dim>>(),
       dt_m(0),
       it_m(0),
       integration_method_m(integration_method),
       solver_m(""),
       lbt_m(lbt),
       isFirstRepartition_m(true),
-      //nt_m(nt),
+      // nt_m(nt),
```
aaadelmann: delete this line

aaadelmann (approved):

src/PartBunch/PartBunch.cpp
aliemen: Done. Removed from the constructor and from its argument list.

PR #396, opened 2026-04-29
Add utf-8 encoding to generated `...ElementPosition.py` file

aliemen:
Otherwise the webview export will shows escaped "`\n`" characters and not show the nice 3D view.


aliemen:
HTML export now works, see e.g. [fodo-short_ElementPositions.html](https://github.com/user-attachments/files/27195291/fodo-short_ElementPositions.html)

aaadelmann (approved):

PR #398, opened 2026-04-29
Fix conic aperture

### LINKED ISSUE #397: Conic aperture scaling starts at zero at element entrance
Current conic aperture handling applies the scale factor as fraction along the element times `f`.
This makes the effective transverse aperture equal to zero at the entrance (fraction = 0), which is physically wrong and leads to artificial particle losses near element entry.

Impact:
- Unexpected losses with conic aperture definitions, including the common case where `f = 1`.
- Inconsistent behavior with the documented meaning of `f`, where f should scale exit aperture relative to the entrance aperture.

Expected behavior:
- Conic scaling should interpolate from 1 at the entrance to f at the exit.
- For `f = 1`, aperture should remain constant along the element.
- Aperture argument parsing should be strict and reject malformed or invalid definitions.
- Conic factor `f` must be strictly positive.

pecalvop:
Closes #397

This pull request introduces improvements to the handling and validation of conic apertures in beamline elements, especially for rectangular and elliptical types, and adds new unit tests to ensure correctness. The main changes include more robust input validation for conic aperture parameters, improved calculation of aperture scaling along the element length, and the introduction of comprehensive unit tests for drift elements.

**Aperture Handling and Validation Improvements:**

* Added input validation for conic aperture scale parameters to ensure they are positive real values, throwing an exception otherwise. This prevents invalid aperture definitions from being accepted. (`src/Elements/OpalElement.cpp`)
* Improved regex patterns for parsing aperture arguments, making them stricter and more robust against malformed input. (`src/Elements/OpalElement.cpp`)

**Aperture Scaling Logic:**

* Updated the calculation of the scaling factor for conic rectangular and elliptical apertures to interpolate smoothly from the start to the end of the element, clamping the interpolation fraction to the [0,1] range for safety. (`src/AbsBeamline/ElementBase.cpp`)

**Testing Enhancements:**

* Added a new unit test file `TestOpalDrift.cpp` with comprehensive tests for drift elements, covering various aperture shapes and conic aperture behaviors, ensuring correct handling of edge cases and equivalence between similar aperture definitions. (`unit_tests/Elements/TestOpalDrift.cpp`)

**Refactoring and code style:**
* Refactored constructors, destructors, and several methods in `ElementBase` and `OpalElement` to single-line definitions for brevity and consistency.
* Reformatted attribute initialization in `OpalElement` for clarity and to reduce code duplication.


src/AbsBeamline/ElementBase.cpp
```diff
@@ -249,9 +235,14 @@ bool ElementBase::isInsideTransverse(const Vector_t<double, 3>& r) const {
     double factor        = 1.0;
     if (aperture_m.first == ApertureType::CONIC_RECTANGULAR
         || aperture_m.first == ApertureType::CONIC_ELLIPTICAL) {
-        Vector_t<double, 3> rRelativeToBegin = getEdgeToBegin().transformTo(r);
-        double fractionLength                = rRelativeToBegin(2) / getElementLength();
-        factor                               = fractionLength * aperture_m.second[2];
+        const double length = getElementLength();
+        if (std::abs(length) > 0.0) {
```
aliemen: If it was negative, `rRelativeToBegin(2) / length` flips sign, the clamp masks it, and you silently get a constant aperture, right? So perhaps using `length > 0` would be safer?

src/Elements/OpalElement.cpp
```diff
@@ -172,9 +150,11 @@ std::pair<ApertureType, std::vector<double> > OpalElement::getApert() const {
...
                 retvalue.second[1] = retvalue.second[0];
                 retvalue.second[2] = std::stod(match[2]);
-            } catch (const std::exception &ex) {
-                throw OpalException("OpalElement::getApert()",
-                                    "could not convert '" + arguments + "' to doubles");
+                validateConicScale(retvalue.second[2], arguments);
+            } catch (const std::exception& ex) {
+                throw OpalException(
+                        "OpalElement::getApert()",
+                        "could not convert '" + arguments + "' to doubles");
```
aliemen: `validateConicScale` can throw a `OpalException`, but `OpalException` [does not inherit from `std::exception`](https://github.com/OPALX-project/OPALX/blob/master/src/Utilities/OpalException.h), meaning I'm not sure if this exception will actually be caught?

unit_tests/Elements/TestOpalDrift.cpp
```diff
@@ -0,0 +1,157 @@
+//
```
aliemen, line 1: Here you could perhaps add (feel free to ignore if you disagree):
- test for the validation path (i.e. that `RECTANGLE(1,1,-1)`, `CIRCLE(1,0)`, `ELLIPSE(1,1,nan)` actually throw. 
- Conic-ellipse test (`ELLIPSE(1,1,2)`). Cheap to add and exercises the conic-elliptical branch in `isInsideTransverse`.
- Test that the regex rejects malformed input (e.g. `RECTANGLE(1,)`, `RECTANGLE(,1)`).

src/Elements/OpalElement.cpp
```diff
@@ -343,29 +326,23 @@ void OpalElement::parse(Statement& stat) {
...
-    if (parent != 0  &&  ! parent->getOpalName().empty()) {
-        if (! getOpalName().empty()) head += ':';
+    if (parent != 0 && !parent->getOpalName().empty()) {
+        if (!getOpalName().empty()) head += ':';
         head += parent->getOpalName();
     }
 
     os << head;
-    os << ';'; // << "JMJdebug OPALElement.cc" ;
+    os << ';';  // << "JMJdebug OPALElement.cc" ;
```
aliemen: What does this comment mean? Perhaps can be removed?

src/Elements/OpalElement.cpp
aliemen: Or is that intended, since those two are separate? In that case it might be clearer to put `validateConicScale` outside of the `try` block?

aliemen (changes requested):
Nice. This PR just fixes the input handling, right? If I understand it correctly, then it doesn't implement any deletion logic yet - in that case, it looks good from my side (except for the small comments).

aaadelmann (approved):
Agree with Alex

aliemen (approved):
Unit/regression tests pass on CPU/GPU (GH200). LGTM

PR #401, opened 2026-04-30
Just a test: minimal `Drift-1` regression test in GitHub runner

aliemen:
This PR restructures and extends the workflows and status check we have to OPALX. We still have required and non-required checks. Required checks _have to pass_ before a merge. This PR will implement the following checks. 

## Formatting Checks
- `Check clang-format`. Checks the formatting of changed files in the PR. **Not required**.

## Compilation
- `CPU Serial`. Compiles serial, debug opalx and unit tests. **Required**.
- `CPU OpenMP`. Compiles openmp, debug opalx and unit tests. **Required**.
- `GPU CUDA`. Compiles CUDA (NVIDIA `AMPERE80`), debug opalx and unit tests. **Required**.
- `GPU HIP`. Compiles HIP (AMD `AMD_GFX90A`), debug opalx and unit tests. **Required**.
- `GPU SYCL`. Compiles SYCL (Intel `INTEL_PVC`), release (memory constraints) opalx and unit tests. **Not required** (since we cannot test it).

## Running Tests
- Unit Tests with the `CPU Serial` build from above.
- Unit Tests with the `CPU OpenMP` build from above.
- A subset of fast regression tests from [OPALX-project/regression-tests-x](https://github.com/OPALX-project/regression-tests-x) with the `CPU Serial` binary. Checks against their `.rt` metrics. Currently:
  - Drift-1-fromfile
  - Drift-4-multi-emit-open
  - FodoCell-fromfile


aliemen:
Cool, looks like we could also:
- Reuse the compiled serial binary to run some simple regression tests 
- Compile for `SYCL` (with some caveats, and IPPLs `FFT.h` needs a few lines)

aliemen:
If we we merge this PR, there are a few things to consider:
- IPPL-framework/IPPL#500 should probably be merged first for the SYCL compilation. 
- The "required" needs to be removed from the legacy checks (branch protection rules), then remove the legacy workflow file.
- After the merge, the branch protection rules should be updated to include the new checks.

aaadelmann:
In the PR describe what this does and what the asumptions are (for example regression tests must be short, only CPU (omp?) is tested)

mohsensadr:
@aliemen very nice. now we're closer to a full ci/cd. I will review this PR closely tomorrow.

Just a comment:

- why "Run clang-format check" is disabled? is it because it won't pass currently?
https://github.com/OPALX-project/OPALX/actions/runs/25370996138/job/74394173000?pr=401

aliemen:
@mohsensadr look at [this line](https://github.com/OPALX-project/OPALX/actions/runs/25370996138/workflow?pr=401#L54). It will only start if the number of changed code files is not 0. In this PR, only the `.yml` files are modified - that's why it's a trivial pass.

It should be the same workdflow file as in master, so e.g. [here](https://github.com/OPALX-project/OPALX/actions/runs/25368175184/job/74384400588?pr=409) it was executed.

mohsensadr:
@aliemen Aren't we overcomplicating (over-designing) ci/cd with downloading binary from dependent runners who compiled the code, and then running tests separately? Or, am I missing something?
Why not just adding a test section after unit test of workflow for each architecture and run the smoke test there?
I see 13 checks, some depend on the other ones.

aliemen:
@mohsensadr There are a few reasons why I did this: 
- That way we can automatically have separate checks in the status checks menu for every test. Meaning, if something fails you immediately see which test actually fails.
- Cleaner separation between running and execution. It's not really feasable for unit tests, but perfect for the OPALX binary.
- Memory constraints. A standard runner has only 14 GB of SSD space. With the SYCL test I already had to use Release, since Debug would generates huge binaries. It probably will not be a problem for serial debug, but having tests, downloaded regression tests and run artifacts might fill it at some point.
- I think the current setup would also make it very easy to reuse the testing matrix for all the other binaries if needed (should that be desired in the future).

So I would argue that separating compilation from the regression matrix is much cleaner and easier to maintain. At the moment, only the regression tests depend on cpu-serial (because it needs to binary). The rest should be independent. Also 4 of the checks are just there to satisfy current branch protection rules (will be removed).

aliemen:
Now that I think about it: I am not sure if testing a custom "subset of regression tests" defined by [a list](https://github.com/OPALX-project/OPALX/blob/79555db49b546b212e65464209c3c2b29a33207d/.github/workflows/cpu-serial.yml#L69) would even work having it in one status check. The nice thing about a [matrix strategy](https://docs.github.com/de/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstrategymatrix) is that it automatically runs each config, but it cannot be nested into another status check. Otherwise, we would probably need to hardcode the tests into the injected python code.

mohsensadr:
Fair enough. It looks good to me.

mohsensadr (approved):

PR #406, opened 2026-05-04
Activate particle IDs in OPALX

### LINKED ISSUE #73: runtime error when saving ids to hdf5
### Summary

runtime error when saving ids to hdf5

### Steps to reproduce

uncomment l461 in  H5PartWrapperForPT.cpp and run DriftTest-0

### What is the current *bug* behavior?

/psi/home/adelmann/opal-x/downloads/opal-x/src/Structure/H5PartWrapperForPT.cpp: [ H5PartWrapperForPT::writeStepData() ]
      ...
      464     Kokkos::deep_copy(idView,idViewDevice);
      465       
      466     for (long unsigned i = 0; i < numLocalParticles; i++)
==>   467         i64buffer[i] = idView(i);
      468     WRITEDATA(Int64, file_m, "id", i64buffer);
      469 
      470     auto binViewDevice  = bunch->getParticleContainer()->Bin.getView();

---
_Migrated from GitLab_
GitLab ID: 11
Source: https://gitlab.psi.ch/OPAL/opal-x/src/-/issues/11
Original state: opened
Original author: adelmann

aliemen:
closes #405 
closes #73 

This PR activates particle IDs in OPALX by explicitly passing the `Kokkos::DefaultExecutionSpace::memory_space` template to every `ParticleContainer` instance (and parent `ParticleBase` definition) and adds a unit tests for the IDs. 

Apart from that I also went ahead and added the IDs back into the H5 writer (as it was before, see #73).

### DIFF
--- a/src/PartBunch/ParticleContainer.hpp
+++ b/src/PartBunch/ParticleContainer.hpp
@@ -60,6 +60,17 @@ template <typename T, unsigned Dim = 3>
 class ParticleContainer
     : public ippl::ParticleBase<
               ippl::ParticleSpatialLayout<T, Dim>, Kokkos::DefaultExecutionSpace::memory_space> {
+    /**
+     * @brief Alias for the `ippl::ParticleBase` specialization this container inherits from.
+     *
+     * The second template argument is a parameter pack of Kokkos view properties forwarded to
+     * the optional `ID` attribute's storage. IPPL gates the `ID` attribute on
+     * `sizeof...(IDProperties) > 0`, so passing any property here turns IDs ON; an empty pack
+     * would leave them disabled. We pass `Kokkos::DefaultExecutionSpace::memory_space` so the
+     * ID view lives in the same space as the rest of the bunch (host or device, matching the
+     * build backend). With IDs enabled, IPPL auto-assigns globally unique `std::int64_t` IDs in
+     * `Base::create()`.
+     */
     using Base = ippl::ParticleBase<
             ippl::ParticleSpatialLayout<T, Dim>, Kokkos::DefaultExecutionSpace::memory_space>;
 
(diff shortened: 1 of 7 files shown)


src/PartBunch/ParticleContainer.hpp
```diff
@@ -57,8 +57,11 @@ using size_type = ippl::detail::size_type;
  * They automatically select the correct underlying storage mode.
  */
 template <typename T, unsigned Dim = 3>
-class ParticleContainer : public ippl::ParticleBase<ippl::ParticleSpatialLayout<T, Dim>> {
-    using Base = ippl::ParticleBase<ippl::ParticleSpatialLayout<T, Dim>>;
+class ParticleContainer
```
aaadelmann, line 60: I do not see why this activates IDs?

aaadelmann (changes requested):

src/PartBunch/ParticleContainer.hpp
aliemen: IDs are activated by the number of template arguments in the variadic pack, see [here](https://github.com/IPPL-framework/ippl/blob/27d11d0b58bc5db6a582f8132c3d6c88285b26bd/src/Particle/ParticleBase.h#L88). 

If it's more than one, the `constexpr` `EnableIDs` is `true` and the templates are passed to the view of the ID. Meaning, the memory space config passed as a template will...
1. implicitly set `EnableIDs == true`,
2. define the memory space where the IDs live.

I set the memory space to `default`, meaning the IDs always live in the same space as the rest of the attributes.

aaadelmann: Please add this as doc to the code.

aaadelmann:
Will approve when the Regression-Tests have passed

aliemen:
Ran the unit tests over lunch: all passed on CPU and GH200. Will add the doc regarding why this activates IDs in a bit.

aaadelmann (approved):

PR #407, opened 2026-05-04
BIG clang-format on everything in the codebase (only codefiles)

aliemen:
> Don't look, just merge. 

For reference, here the command that formats the whole codebase (if you want to format your branch yourself instead of merging master - which might generate much less conflicts):
```bash
find src unit_tests -type f \( -name "*.cpp" -o -name "*.hpp" -o -name "*.h" -o -name "*.tpp" \) -print0 | xargs -0 clang-format -i --style=file
```
This command needs to be called from the OPALX root.


aaadelmann (approved):

aliemen:
Had to do some local git-gymnastics, because somehow GitHub left the conflict markers in the code after clicking "Update Branch"??? But it should be fine now. 

Will run the regression tests and then merge.

aaadelmann:
and how Do I fix my branch? Merge in master is probable the best !?

aliemen:
My recommendation is the following. First run (after thir PR goes through):
```bash
git merge -X ignore-all-space master
```
...generates barely any conflicts, since formatting is more or less just whitespace changes. 

After that, run `clang-format` on your branch again. From the code root something like:
```bash
find src unit_tests -type f \( -name "*.cpp" -o -name "*.hpp" -o -name "*.h" -o -name "*.tpp" \) -print0 | xargs -0 clang-format -i --style=file
```

I tested in on your `391-...` branch, only gave two small trivial conflicts. 

Edit: I'll post it in the OPALX chat, since others may run into the same issue.

aliemen:
Update: regression tests and unit tests look good on CPU + GPU (GH200). Will merge in 15min after the CI completed.

Issue #395, opened 2026-04-28
Restore and migrate `Monitor` element from OPAL to OPALX


MichZampetakis:
Following the migration series of the RF cavities from OPAL to OPALX, a necessary addition for SwissFEL simulations is the migration of the `Monitor` element.

The `Monitor` element appears to already exist in OPALX, but its implementation is currently disabled. In issue [#120](https://github.com/OPALX-project/OPALX/issues/120), the functions in `AbsBeamline/Monitor.cpp` were commented out after compile errors were encountered during the CMake rewrite. 

The goal of this issue is to bring the `Monitor` element back to a working state in OPALX, while keeping the implementation compatible with both CPU and GPU execution. This is needed to support SwissFEL use cases and to continue the broader OPAL → OPALX element migration effort. The existing blocker/bug is already tracked in [#120](https://github.com/OPALX-project/OPALX/issues/120), which was opened on November 10, 2025. 

**Expected outcome**

* Re-enable the `Monitor` element implementation in OPALX
* Resolve the compilation issues currently tracked in #120
* Ensure the implementation is compatible with both CPU and GPU execution
* Preserve, as far as practical, the expected OPAL behavior/interface for existing workflows
* Add at least one validation test or example covering `Monitor` functionality

MichZampetakis:
Notes:
- `Monitor` compiles now but there are issues when a `Monitor` is used in a simulation. Segmetation fault when trying to get particle hits or save file.
- Need to find a way to get particle IDs from the Particle container.
- Issues with `src/Structure/LossDataSink.cpp`. In particular `LossDataSink::saveH5()` and `LossDataSink::hasNoParticlesToDump()` (so far)

MichZampetakis:
Strange behavior for Spatial VS Temporal Monitor? Simulation for 10000 macroparticles and energy 6.6 MeV.
```
SPATIAL  -> loss-particle convention  -> ENERGY ≈ 0.6179094
                                         TotalMass: 478.41068752
TEMPORAL -> bunch-particle convention -> ENERGY ≈ 6.6
                                         TotalMass: 5109.9895
```
Same `gamma` in both cases.

PR #409, opened 2026-05-05
Add `InvalidMask` as particle attribute

aliemen:
Add the `InvalidMask` attribute to the `ParticleContainer`. 

The mask isn't integrated yet and doesn't do anything, but can already be used for the developement of #395 and `LossDataSink`. How this will be used is explained in #404, but I'll make a separate PR for that later.

### DIFF
--- a/src/PartBunch/ParticleContainer.hpp
+++ b/src/PartBunch/ParticleContainer.hpp
@@ -139,7 +139,7 @@ public:
     typename Base::particle_position_type B;
 
     /// particle deletion mask (indicates which particles are deleted every timestep)
-    ippl::ParticleAttrib<bool> invalidMask;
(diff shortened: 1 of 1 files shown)


aaadelmann (approved):

aliemen:
Info: don't delete the branch - integration of the attribute will be implemented on the same branch

aaadelmann:
We have the rule that PR which are closed should also be deleted. Please do so otherwise we will end with a mess. This is a super small PR, you do not need to save this one.

aliemen:
OK will do that. (_the issue that's linked to the branch, #404, deliberately was not resolved by this PR, so I'll rewrite the old issue to match this PR, then close it and make a new one_)

PR #410, opened 2026-05-06
Unify particle deletion API

aliemen:
closes #404

This PR unifies the particle deletion. EVERY action that results in the deletion of particles has to go through `ParticleContainer::deleteInvalidParticles`. 

If a process decides it wants to delete particles, it would then set the corresponding index of the `pc->InvalidMask` attribute to `true`. Then, the tracker loop in `ParallelTracker::execute` will call `deleteInvalidParticles` for every container and reset the `InvalidMask` attribute. 

The Decay and boundpDestroy now no longer call IPPL deletion directly.
- `ParticleContainer::markParticlesOutside` now goes through the particles and simply sets `InvalidMask[i] = true` for all particles outside of the $n$-sigma range of the mean.
- The `Decay` previously used its own `invalid` mask, this is not changed (I only renamed to `decayed` instead of `invalid` to avoid confusion). However, at the end it will calculate something like `InvalidMask = InvalidMask || decayed` to mark decayed particles for deletion.

At the moment, `deleteInvalidParticles` is called twice in the tracker, after global processes and after `markParticlesOutside`. This preserves deletion timing from before.


aliemen:
Update: all unit tests and regression tests pass on CPU and GPU (GH200). Even the not-yet-merged decay regression test looks good.

aaadelmann (approved):

PR #414, opened 2026-05-08
Proof OPALX for Kokkos 5 and IPPLs PIF PR

aliemen:
closes #412 

This PR introduces two things:
- Fixes that are necessary on the OPALX side to run with IPPL after the IPPL-framework/IPPL#501 PR went through. 
  - Use `.size()` instead of `pc->R.getView().extent(0)`, since this will now return a tailored subview.
  - `getView` returns a temporary subview: needs to bind to an lvalue, so passing "`R.getView()`" directly as a reference to a function doesn't work anymore. Shallow copies are better anyways!
- Some API depracations that were introduced with Kokkos 5. 
  - `DualView` direct view access is now private, need to use the getter. 
  - `atomic_increment` was replaced by `atomic_inc`.
  - Some default initialization does NOT default to `0` anymore, which surfaced some bugs in `MultipoleT`. @Jonathan-A-Thompson could you please check that these were _actually_ bugs and I fixed them correctly? 

Apart from that I fixed some logging levels. Apparently the level is reset after `endl` was called.


Jonathan-A-Thompson (approved):

aliemen:
Unit and regression tests work on CPU and GPU (GH200). Tested with Kokkos 4 and IPPL master as well as Kokkos 5 and IPPL's new PIF PR.

src/Fields/FM2DMagnetoStatic.cpp
```diff
@@ -110,7 +110,7 @@ FM2DMagnetoStatic::FM2DMagnetoStatic(std::string aFilename) : Fieldmap(aFilename
 FM2DMagnetoStatic::~FM2DMagnetoStatic() { freeMap(); }
 
 void FM2DMagnetoStatic::readMap() {
-    if (FieldstrengthBz_m.h_view.data() == nullptr) {
+    if (FieldstrengthBz_m.extent(0) == 0) {
```
aaadelmann, line 113: is there is no size()

aaadelmann (approved):

src/Fields/FM2DMagnetoStatic.cpp
aliemen: Sadly no, for Kokkos (Dual)Views `size()` [doesn't exist](https://kokkos.org/kokkos-core-wiki/API/containers/DualView.html#_CPPv4I0ENK8DualView6extentENSt9enable_ifINSt11is_integralI5iTypeE5valueE6size_tE4typeERK5iType).

aliemen:
Update: Kokkos 5 requires CUDA 12.2 or newer. However, Gwendolen only has openmpi compiled against cuda 12.1 (or at least I didn't find any other openmpi). 

To use Kokkos 5 on Gwendolen, we either need to compile it ourselves or ideally someone adds a module with e.g. gcc $\geq$ 13, some openmpi $\geq$ 4.1.6 and cuda $\geq 12.2$.  

Something like the following toolchain would be nice:
- gcc 15
- openmpi 5.X
- cuda 12.9.X

Edit: there is nothing anymore that I can change here in OPALX regarding Gwendolen compatibility - that's a toolchain issue (in the case of Kokkos 5) or a IPPL PIF PR problem in the case of cuda12.1 and gcc12. So I'll just merge this PR.

After the merge, OPALX (with current IPPL master) works for Kokkos 5 and works with Kokkos 4 & 5 with the PIF branch on Daint for newer toolchains.

PR #415, opened 2026-05-09
Fix binary repartition exception call (some multirank tests) and introduce `STEPINFOFQ` option

aliemen:
While trying out multi rank regression tests, I noticed that they fail, since I previously added an exception should the binary repartition routine be called. However, since some tests (like `FODO`) called it anyway (the function calls are wired into OPALX, just no-ops), I fixed it by writing a generic warning to the console and not a hard exception. 

## Summary
Disable the currently unwired binary load-balancer repartition path while preserving `REPARTFREQ` input "compatibility". Repartition requests now log and skip safely instead of calling incomplete ORB/multi-container logic.

Add a new `STEPINFOFQ` option to control per-step tracking status output, including support for `0` to silence those lines. Beam statistics dump logging is also lowered in verbosity.

## Changes
- Disable scheduled binary repartitioning and guard against zero repartition frequency.
- Convert `PartBunch::do_binaryRepart()` from an exception path to a logged no-op.
- Add `Options::stepInfoFreq` and expose it through `OPTION, STEPINFOFQ`.
- Gate tracker step-status lines on `STEPINFOFQ`.

> FYI: I did the second change with `STEPINFOFQ`, since huge simulations with thousands of steps (`AWAGun`) then have unnecessarily verbose logs. Old OPAL also has a mechanism that automatically only prints every 100th step for `AWAGun`. 

Edit: also did a little fix to the workflows. With multiple labels, they always triggered multiple runs for each config, then deleted the old ones and only kept the newest. That's unnecessary, now they are only triggered once for every action by giving them their own concurrency group.

### DIFF
--- a/src/Algorithms/ParallelTracker.cpp
+++ b/src/Algorithms/ParallelTracker.cpp
@@ -641,7 +641,7 @@ void ParallelTracker::computeSpaceChargeFields(unsigned long long step) {
     // TODO: itsBunch_m->boundp() not implemented yet.
     // itsBunch_m->boundp();
 
-    if (step % repartFreq_m + 1 == repartFreq_m) {
+    if (repartFreq_m > 0 && step % repartFreq_m + 1 == repartFreq_m) {
         doBinaryRepartition();
         m << level4 << "Binary repartition done." << endl;
     }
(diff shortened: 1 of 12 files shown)


aliemen:
Update: unit and regression tests pass on CPU and GPU (Daint GH200). In both cases now also multirank.

src/Algorithms/ParallelTracker.cpp
```diff
@@ -1318,15 +1317,26 @@ void ParallelTracker::setOptionalVariables() {
     */
     Inform m("ParallelTracker::setOptionalVariables");
 
-    // there is no point to do repartitioning with one node
+    repartFreq_m = 0;
+
+    // The ORB load-balancer path is not currently wired to the unified
```
aaadelmann: Do not understand the comment

aaadelmann (changes requested):

src/Algorithms/ParallelTracker.cpp
aliemen: The comment is supposed to hint at the fact that the ORB (which the Load Balancer is based on) is inherently a [per container action](https://github.com/IPPL-framework/ippl/blob/c2ec8aa48e534d8ced02333dff4c7bb7c6e86d3a/src/Decomposition/OrthogonalRecursiveBisection.h#L55). However, our `PartBunch` currently only passes the primary container to `LoadBalander.cpp`. At the moment, binary repart is a no-op anyways, but now there is a comment that hints at a problem that needs to be adressed when activating it later on.

aaadelmann: Ok then please write it so !

aliemen: Done, the comment should now be clearer.

aaadelmann (approved):

PR #418, opened 2026-05-12
Add `Astra1DMagnetoStatic` fieldmap support

MichZampetakis:
Closes https://github.com/OPALX-project/OPALX/issues/417

## Summary

This PR adds support for the `Astra1DMagnetoStatic` fieldmap in OPAL-X.

The implementation follows the same modernization approach used for `Astra1DDynamic`:

- migrates the fieldmap from the old `_Fieldmap` interface to the current `Fieldmap` interface;
- stores Fourier coefficients in a `Kokkos::DualView`;
- provides a device-safe `computeField(...)` implementation;
- adds `applyField(...)` for particle-container field application;
- replaces the old GSL FFT preprocessing path with explicit Fourier coefficient computation;
- keeps GSL spline interpolation for preprocessing the input fieldmap onto an equidistant grid;
- adds validation for invalid fieldmaps, too few samples, and zero-amplitude maps.

## Motivation

This fieldmap is needed for SwissFEL-related simulations and appears to be used for some solenoid fieldmaps.

The work is similar in spirit to the recent `Astra1DDynamic` / `TravelingWave` migration.

Related PR:
- https://github.com/OPALX-project/OPALX/pull/376

## Tests

A new unit test target for `Astra1DMagnetoStatic` was added.

A follow-up regression test will be created using the SwissFEL solenoids.


MichZampetakis:
To test the `Astra1DMagnetoStatic` Fieldmap, I compared OPAL with OPALX over the following slightly modified SwissFEL line: 
`Line = (Solenoid)` + `(TW1, Solenoid, Solenoid, Solenoid, Solenoid)` + `(TW2, Solenoid, Solenoid, Solenoid, Solenoid);`

Multiplot_comparison

I believe it seems good.

Input files: [Solenoids.zip](https://github.com/user-attachments/files/27708460/Solenoids.zip)

> I tried to include the RF Gun as well but it seems that the SwissFEL RF gun used an AstraDynamic Fieldmap that probably is not supported yet. I will check in a separate issue.


**Should I add an explicit `Solenoid` regression test using the `Astra1DMagnetoStatic` field map in `regression-tests-x`?**

src/Fields/Astra1DMagnetoStatic.cpp
```diff
@@ -0,0 +1,357 @@
...
+#include "PartBunch/PartBunch.h"
+#include "Physics/Physics.h"
+#include "Utilities/GSLSpline.h"
+#include "Utilities/GeneralOpalException.h"
+#include "Utilities/Util.h"
+
+#include <algorithm>
+#include <cmath>
+#include <fstream>
+#include <ios>
```
pecalvop: Included header ios is not used directly

src/Fields/Astra1DMagnetoStatic.h
```diff
@@ -0,0 +1,95 @@
+#ifndef OPALX_AstraFIELDMAP1DMAGNETOSTATIC_HH
+#define OPALX_AstraFIELDMAP1DMAGNETOSTATIC_HH
+
+#include "Fields/Fieldmap.h"
+#include "Physics/Physics.h"
+
+#include <Kokkos_Core.hpp>
+#include <Kokkos_DualView.hpp>
+
```
pecalvop, line 11: I miss `#include <memory>`

src/Fields/Fieldmap.cpp
pecalvop, line 42: Some standard headers should be included for the completeness of the code: `sstream, iomanip, cstring, cctype`. Additionaly, Options.h does not appear to be used

src/Fields/Astra1DMagnetoStatic.cpp
MichZampetakis: Removed `#include <ios>`, thanks!

src/Fields/Astra1DMagnetoStatic.h
MichZampetakis: Added `#include <memory>`

src/Fields/Fieldmap.cpp
MichZampetakis: Removed: `#include "Utilities/Options.h"`
Added: 
- `#include <cctype>`
- `#include <cstring>`
- `#include <iomanip>`
- `#include <sstream>`

Thanks.

pecalvop (approved):

MichZampetakis:
Same test on Gwendolen using: `gcc/12.4.0`, `openmpi/4.1.6_slurm`, `cmake/3.25.2`.
4 GPUs were used on a single node. Results seem good.

image

MichZampetakis:
Not sure why the Drift Smoke Test failed.

aaadelmann:
I would argue with 10'000 particles achieving 1e-4 particles is statistically not possible 1/\sqrt{1e4} = 1/100 = 1% and of course this is the best case. In the "smoke test" it is 1e-3 required.

aliemen:
@MichZampetakis: `Drift-4-multi-emit-open` fails, because it's already picking up my re-do. I'll merge #423 asap and then your status checks should pass, too.

aaadelmann:
@aliemen if this is a from file distribution then ignore my comment above otherwise I still think 1e-3 is too tight.

PR #421, opened 2026-05-13
Support `Astra1DDynamic` fieldmaps in `RFCavity`

MichZampetakis:
This PR extends the particle-container application path of `RFCavity` to support `Astra1DDynamic` fieldmaps in addition to the existing `FM2DDynamic` path.

Closes: https://github.com/OPALX-project/OPALX/issues/420

### Motivation

`RFCavity::apply(pc)` currently supports only `FM2DDynamic` fieldmaps. However, some cases, such as the SwissFEL RF gun, require `Astra1DDynamic` fieldmaps. This PR adds the missing support so these cases can be migrated to OPALX.

### Changes

* Added an `Astra1DDynamic::applyRFField(...)` path for RF-scaled particle application
* Added a device-callable `Astra1DDynamic::computeRFField(...)` helper
* Updated `RFCavity::apply(pc)` to dispatch to either:

  * `FM2DDynamic::applyRFField(...)`
  * `Astra1DDynamic::applyRFField(...)`
* Kept the same RF scaling convention as the existing `FM2DDynamic` path:
  * electric field scaled with `scale * cos(phi)`
  * magnetic field scaled with `-scale * sin(phi)`

### Tests

* Extended `Astra1DDynamic` unit tests for the new RF-cavity helper logic
* Verified that the existing `RFCavity` unit tests remain unchanged
* Test performed using the SwissFEL RF gun seems good
Multiplot_comparison

### Notes
* Full SwissFEL regression test will follow

### DIFF
--- a/src/AbsBeamline/RFCavity.cpp
+++ b/src/AbsBeamline/RFCavity.cpp
@@ -10,6 +10,7 @@
 #include <filesystem>
 #include "AbsBeamline/BeamlineVisitor.h"
 #include "Component.h"
+#include "Fields/Astra1DDynamic.h"
 #include "Fields/FM2DDynamic.h"
 #include "Fields/Fieldmap.h"
 #include "PartBunch/PartBunch.h"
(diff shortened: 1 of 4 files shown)


aaadelmann (approved):
Just make sure the tests also run on more than one rank and A100 (minimum)

MichZampetakis:
Same test on Gwendolen using:`gcc/12.4.0`, `openmpi/4.1.6_slurm`, `cmake/3.25.2`. 
4 GPUs were used on a single node. Results seem good.

image

When I used `gcc/15.2.0` `openmpi/5.0.10_slurm`, the .stat file consided of 4 times the data because probably each task/GPU writes in the .stat file.

Issue #306, opened 2026-03-22
Drifts with space charge give wrong results on Daint's GH200


aliemen:
Running [`Drift-3-open`](https://github.com/OPALX-project/regression-tests-x/tree/master/RegressionTests/Drift-3-open) or [`Drift-3-periodic`](https://github.com/OPALX-project/regression-tests-x/tree/master/RegressionTests/Drift-3-periodic) in Debug mode on GH200 results in totally wrong energies. 

## Open Solver
```text
OPAL-X> * 14:44:29 Step      0 at 29.655 [um], t= 50.000 [ps], E=16444.089 [GeV]
OPAL-X[2]> * Wrote beam statistics.
OPAL-X> * 14:44:30 Step      1 at 59.310 [um], t= 100.000 [ps], E=16601.519 [GeV]
OPAL-X[2]> * Wrote beam statistics.
...
```
Here, we would expect the following.
```text
OPAL-X> * 14:21:12 Step      0 at 29.655 [um], t= 50.000 [ps], E=1.059 [eV]
OPAL-X[2]> * Wrote beam statistics.
OPAL-X> * 14:21:12 Step      1 at 59.310 [um], t= 100.000 [ps], E=1.232 [eV]
OPAL-X[2]> * Wrote beam statistics.
```

## Periodic Solver
```text
OPAL-X> * 14:49:51 Step      0 at 29.655 [um], t= 50.000 [ps], E=6.471 [MeV]
OPAL-X[2]> * Wrote beam statistics.
OPAL-X> * 14:49:51 Step      1 at 59.310 [um], t= 100.000 [ps], E=6.495 [MeV]
OPAL-X[2]> * Wrote beam statistics.
```
Here we would expect the following.
```text
OPAL-X> * 14:57:30 Step      0 at 29.655 [um], t= 50.000 [ps], E=1.046 [eV]
OPAL-X[2]> * Wrote beam statistics.
OPAL-X> * 14:57:30 Step      1 at 59.310 [um], t= 100.000 [ps], E=1.184 [eV]
OPAL-X[2]> * Wrote beam statistics.
```

---
This was brought to my attention by @rammann. I will perhaps try to look into this on Monday. I suspect that it has something to do with the solvers again, but hopefully not as bad as the previous GH200 bug... **Note that everything works perfectly fine in release mode.** Since regression tests didn't show anything (and they are run in debug if I'm not wrong), it's probably just an issue on GH200.

aliemen:
[...]

PR #423, opened 2026-05-15
New `OpalFlatTop` and reproducing `AWAGun`

### LINKED ISSUE #393: `FlatTop` distribution doesn't match old OPAL
OPAL pre samples `FlatTop` in time and then moves the whole histogram step by step into the simulation. Here in OPALX we opted for a more flexible solution: sample particles at runtime every timestep. However, there are a few problems with this solution:
1. OPAL starts the simulation (which would be "`t=0`") with half of the distribution already in the simulation. That way, they don't run into problems of statistical inaccuracies during the early emission phase when OPALX has only a few active particles. This might lead to a situation where initially the the number of gridcells in the solver far exceeds the number of particles. 
2. The number of globally emitted particles is calculated every timestep as an integer and of course scales anit-proportional to the timestepsize. Meaning, small timestepsizes (very accurate simulations) lead to many accumulated integer casting errors every timestep. The results is missing particles at the end of the emission phase. 

Here a comparison plot that shows both issues:

Image

One can see how initial conditions at `t=0` are different and how OPALX ends with slightly less particles. Since simulations like [`AWAGun-1`](https://github.com/OPALX-project/regression-tests/blob/master/RegressionTests/AWAGun-1/AWAGun-1.in) are very sensitive to differences in initial conditions, we run into problems. 

I don't have an exact solution for this yet, but three things need to be considered:
- Keep runtime emission, since having just a shift in `t` is not a problem. On the contrary, this should make OPALX more accurate since it actually simulates the full emission phase more faithfully. However, results in `s` should be reproducible.
- Think about "topping" up missing particles that are discarded due to integer casting.
- Think about what to do with the solver when there not enough particles. A possibility would be to calculate the full $N^2$ contributions (not a problem with few particles) or to simply disregard the solver or use less grid points.

aliemen:
closes #393

This PR reproduces the results from old OPAL's `AWAGun` by introducing a new distribution type, `TYPE=OPALFLATTOP` as well as the timestepsize and global time handling from old OPAL for emitted distributions. An explanation oh what was added/changed with respect to the distribution (mainly `OpalFlatTop`) can be seen [here in the issue](https://github.com/OPALX-project/OPALX/issues/393#issuecomment-4460210297). 

It's quite a big PR, but most of the changes are from the new `OpalFlatTop`/`EmittedFromFile` files and their unit tests (and those are quite similar to `FlatTop` and `FromFile`). 

## Distribution changes (explained in more detail in the linked issue)

- Adds `OPALFLATTOP` and `EMITTEDFROMFILE` as valid `DISTRIBUTION, TYPE=...` values, including predefined string support and normal `Distribution` setup/printing paths.
- Adds both new distribution implementations to the build/install lists and wires them into `TrackRun::setupDistributionsAndSamplers()` so the tracker can instantiate them from input files.
- Adds `EMISSIONSTEPS` to `DISTRIBUTION` so OPAL-like emitted distributions can request a finer emission timestep while emission is still active.
- Extends `SamplingBase` with optional hooks for:
  - tracker time shift before emission,
  - initial reference momentum for empty emitting containers,
  - sampler-controlled emission timestep.
- Updates `ParallelTracker` to consume those hooks:
  - initializes empty emitting containers from sampler-provided reference momentum when available,
  - applies the global emission time shift before tracking,
  - selects sampler emission dt while emitted sources are active,
  - resets per-particle `dt` after the full step so fractional emission timing does not leak into later steps.
- Treats `EMITTEDFROMFILE` like `FROMFILE` for beam-energy validation, since particle momenta come from the file and should not be silently overridden by `BEAM` energy settings.
- Initializes newly created particle slots more completely (`InvalidMask`, `dt`, `Phi`, `Bin`, `E`, `B`) so late-emitted particles start from clean state. _I am thinking about removing these again, but they shouldn't hurt too much._

## Solver related changes

- Adds `BINNING, ADAPTIVEBINNING=FALSE` to preserve uniform `MAXBINS` bins for OPAL comparison runs. Basically a way to keep uniform binning. 
- Adjusts binned self-field handling for early-emission cases: global one-particle no-self-field guard, safer direct Kokkos field initialization/scaling, mean-per-particle gamma, all-local scatter bypass, and GH200-safe explicit CIC scatter/host-staged halo path (please read the note below!).
- Fixes shifted Green's image-charge sign handling by avoiding the extra rho negation before the shifted solve.

> Note: In order to run `Release` on Daint (GH200), I replace **all** implicit field operations that are based on expression templates (like `rho = rho / coupling`) with explicit `Kokkos::parallel_for`. Apart from that, I replaced the internal `field->accumulateHalo()` calls wherever necessary with my own copy-to-host style implementation. That is potentially not very efficient, since it doesn't use GPU-aware MPI. However, it works fine on Release GH200. I will add another comment in #306 or in a new issue later that notes that this implementation should be removed at some point. 

## Tests/Regression

- Registers unit tests for `OpalFlatTop` and `EmittedFromFile`.
- Adds coverage for `EMITTEDFROMFILE` predefined string parsing.
- Adds tests for `ADAPTIVEBINNING` defaults/getters and uniform-bin preservation in the binned solver.

Unit tests and regression tests pass as expected single- and multirank on CPU and on GPU (GH200). However, because of some solver and FlatTop modifications, `Dist-flattop` and `Drift-4-multi-emit-open` need to be re-done. 

## `AWAGun` plot

Using `1E4` particles and a `16^3` grid with adaptive binning in OPALX (`DESIREDWIDTH=0.2`) and uniform 10 bins over the whole OPAL run, we get:

comparison

There is some difference in the emittance. However, when we compare a `32^3` run with `1E6` particles, we already get better results (here I now used 10 uniform bins in both runs):

comparison

That should support my hypothesis: the remaining differences are due to statistical effects in the solver as well as the number of particles in the simulation. 

> Also note worthy, the GH200 in the last plot was done on 16 ranks (4 nodes) on Daint in Release mode and completed in $475s$. 

Another interesting note: using 10 uniform bins instead of the adaptive binning algorithm cuts executiong for the `16^3` and `1E4` particles on my Mac down to `90s` from `306s`, but gives similar accuracy:

comparison

This should show how well adaptive binning works in this context.


aliemen:
Here some graphs from a bigger simulation. 
- Input file: [AWAGun-1.in](https://github.com/user-attachments/files/27851353/AWAGun-1.in.txt).
- Log file: [opalxg_3435810.out](https://github.com/user-attachments/files/27851359/opalxg_3435810.out.txt).
- OPAL input file for comparison: [opal-AWAGun-1.in](https://github.com/user-attachments/files/27851361/opal-AWAGun-1.in.txt) (ran in `16182s` on 8 ranks on Mac).
- OPAL log file: [opal-AWAGun-1.out](https://github.com/user-attachments/files/27851513/opal-AWAGun-1.out.txt).


comparison

For OPAL I used $128^3$, for OPALX I ran $512^3$ each with `8ppc`. The same simulation on 8 GH200 nodes (32 ranks) resulted in roughly $5 \frac{\mathrm{s}}{\mathrm{timestep}}$. On 128 ranks I got $0.97 \frac{\mathrm{s}}{\mathrm{timestep}}$. That's a `5.1x` speedup for 4x the node count. It's not completely comparable though: the 32 ranks job reached its timelimit after 3000 steps, so the full 128 ranks run had much more timesteps in the later half of the simulation where the number of bins ($\equiv$ solves per timestep) is much lower.

> Note that OPAL only bins for `~200` timesteps, so it is expected to see a slight accumulated difference at the end of the drift.

src/Distribution/OpalFlatTop.h
```diff
@@ -0,0 +1,110 @@
+/**
```
aaadelmann, line 1: In general the new header files are missing some documentation

aaadelmann (changes requested):

src/Distribution/OpalFlatTop.h
aliemen: Good point, I added it. They are quite similar to `FlatTop` and `FromFile`.

aliemen:
After updating the tests OPALX-project/regression-tests-x#36, all of them pass on CPU and GPU (GH200 Daint and A100 on Gwendolen). Each (Mac and Daint) tested also with 2 ranks.

aaadelmann (approved):

PR #427, opened 2026-05-20
sbend and rbend with some examples - Part 1

### LINKED ISSUE #391: Placement of SBEND and RBEND - Analytic Fields *including fringe fields)
## Summary

  The straight-element placement redesign is now in place, but `SBEND` / `RBEND` placement is still missing from the active OPALX code path.

  Current status:
  - `SBEND`, `RBEND`, and `RBEND3D` still exist in `ElementType`
  - `OpalBeamline::compute3DLattice()` still contains bend-specific branches
  - but the actual bend logic is stubbed out and all bend-specific quantities are replaced by zeros

  This issue is about restoring bend placement for the **analytic-field** case first, including **fringe fields**.
   `RBEND3D` and field-map-driven bend placement will follow later.

  #
  ## Motivation

The current placement model already distinguishes:
  - nominal body extent
  - field-support extent

  and already handles:
  - explicit 3D placement
  - overlap bookkeeping through active element sets
  - traced reference-path occupancy

  ## Current gaps

  In particular, the bend-specific placement/export path still expects APIs like:
  - `getChordLength()`
  - `getBendAngle()`
  - `getEntranceAngle()`
  - `getRotationAboutZ()`
  - `getDesignPath()`

These functions have to be converted to the new language. 

## Scope

  For the **analytic field model**, including:
  - body geometry
  - curved reference path
  - entrance/exit edge angles
  - fringe-field extent

  # Design target

  - nominal body placement
  - entry / body / exit ports
  - curved local reference path
  - field-support extent, including fringe fields
  - reference-order occupancy derived from field support

  For bends, the actual traced reference path is still determined by integrating the reference particle through the active fields. The placement model should provide the curved geometry and field-support description needed for that integration.

  ## Proposed first implementation

  ### 1. Recover the old bend wrappers from history

  The deleted wrapper code is still available in git history, for example around commit `34805a452`.

  Relevant files in history include:
  - `src/Elements/OpalBend.h`
  - `src/Elements/OpalSBend.h`
  - `src/Elements/OpalSBend.cpp`
  - `src/Elements/OpalRBend.h`
  - `src/Elements/OpalRBend.cpp`
  - `src/Elements/OpalRBend3D.h`
  - `src/Elements/OpalRBend3D.cpp`
  - `src/Elements/OpalSBend3D.h`
  - `src/Elements/OpalSBend3D.cpp`

  These should be used as reference material, not copied back mechanically.

  ### 2. Reintroduce an explicit bend element contract

  We likely need a bend-side API that exposes at least:
  - bend angle
  - chord length and/or arc length
  - entry angle
  - exit angle
  - optional rotation about local z
  - nominal curved body path
  - field-support interval including fringe fields

  This may be a revived `BendBase`-like abstraction or a new equivalent.

  ### 3. Define bend placement in the new vocabulary

  For bends, the placement model should provide:
  - `entry`
  - `body`
  - `exit`
  - curved local reference path
  - placed support/body geometry
  - field-support extent

  This should integrate with:
  - `PlacedElement`
  - `ReferencePathModel`
  - `OrbitThreader`
  - `MeshGenerator`

  ### 4. Handle fringe fields explicitly

  Fringe fields should not be hidden inside the nominal body length.

  They should contribute to:
  - field-support extent
  - overlap / occupancy intervals
  - traced reference path

  This is consistent with the current body-vs-field-support semantics already introduced for straight field-map-backed elements.

# From #390 

- We have several classes in the BeamlineGeometry folder that are lightweight wrappers of CoordinateSystemTrafoand all functionally the same: Misalignment, PlacedPose, and SupportPlacement. Maybe in the future we can unify these?
- The Bend classes reimplement the Multipole functionality, as addressed in the future we should try to wrap the Multipole functionality similarly to the way it was done for the Quadrupole element.
- Enumerate the Port types instead of using a std::string

  ## Acceptance criteria

  - `SBEND` and `RBEND` can be placed correctly in 3D using the new placement model
  - `compute3DLattice()` no longer uses stubbed bend branches
  - `save3DLattice()` can export meaningful curved bend geometry/path
  - entrance/exit edge angles are represented correctly
  - fringe fields contribute to field support and occupancy
  - at least one regression test covers bend placement geometry
  - at least one regression test covers occupancy / overlap behavior for a bend with fringe fields
  - the Manual / placement note is updated accordingly

aaadelmann:
## Implementation Summary

This PR implements the first analytic bend-placement layer for OPALX, focused on `SBEND`/`RBEND` with fringe/edge support at the analytic-field level.

The main implementation change is that bend tracking no longer relies only on a generic local element frame. Analytic bends are handled through bend-specific field-local coordinates and tracking slices. In `ParallelTracker`, `SBEND`/`RBEND` elements are detected explicitly, transformed through the nominal entry pose plus misalignment, and applied slice-by-slice via `BendBase::buildTrackingSlices()` and `BendBase::applySlice()`.

Reference-particle field evaluation was also updated to use the bend field-local frame consistently. For bends, the reference momentum and fields are rotated between entry-Cartesian and field-local coordinates before/after `applyToReferenceParticle()`. This keeps the tracker and `OrbitThreader` aligned in how analytic bend fields are evaluated.

The branch also keeps the placement model separated from legacy ordering concepts. Element placement is represented through the placed-element transforms, while legacy `ELEMEDGE`-style action ranges are retained only as compatibility/reporting information. Chicane and bend examples now use explicit `X`, `Y`, `Z` placement formulas.

### Validation added or exercised in this branch includes:
- unit coverage for analytic `SBEND` geometry and pole-face/edge-angle behavior
- beamline placement tests for entry/body/exit placement consistency
- monitor placement behavior for zero-length monitor usage at element exits
- single-`SBEND` sandbox examples
- C-type four-bend chicane sandbox example with visualisation
- R56 check for the chicane using the generated H5 output

[Current R56 regression ](https://opalx-project.github.io/PhysicsManual/sections/elements/#example-c-type-chicane)
test and results

```text
transport R56 = -4.311403722785e-02 m
small-angle expected R56: -4.333333333333e-02 m
difference: 2.192961054842e-04 m
```

## Implemented
| Area | Implemented on this branch | Status |
|---|---|---|
| Placement core | `PlacementPose` / `PlacedElement` bridge with nominal body transform, entry/exit port transforms, and misalignment separation | Sound internal model |
| Explicit placement | `X,Y,Z,THETA,PHI,PSI` are accepted for explicitly placed elements | Works, but pose meaning is implicit |
| Analytic bends | `SBEND` / `RBEND` can be placed without `ELEMEDGE` when explicit 3D placement is supplied | Implemented |
| Legacy compatibility | `ELEMEDGE` path remains available for old-style entry-edge placement | Preserved |
| SBEND analytic tests | Unit-level analytic SBEND cases, including bend angle and pole-face behavior | Good start |
| Sandbox SBEND tests | `test-sbend-1`, `test-sbend-2`, with README/test reasoning and bend geometry figures | Useful manual/regression basis is only documented in the [Physics manual](https://opalx-project.github.io/PhysicsManual/sections/elements/#code-to-code-and-analytic-comparisons) |
| C-type chicane sandbox | `test-chicane-1` with symbolic placement formulas, generated placement plot, and `R56` extraction script | Good benchmark seed |
| Physics Manual | Chicane example added to Ch. 3.1.8 with formulas, figure, input, scripts, and references | Integrated |
| Diagnostics | Port-continuity diagnostics and placement reporting restored/kept | Useful, but needs stronger checks |

## Testing

Check [tests that will] be elevated to Regression Tests soon](https://github.com/OPALX-project/OPALX/blob/391-placement-of-sbend-and-rbend-analytic-fields-including-fringe-fields/sandbox/BenTest/bend-tests.pdf)

## Missing / Follow-Up (next PR)

Remaining gaps are intentionally left for follow-up work:

- introduce an explicit POSING/pose-convention attribute instead of relying on implicit convention
- expand regression coverage beyond the current analytic bend/chicane cases
- extend the placement model to field-map bends later
- tighten documentation around all OPAL-X frames and transform directions
- add CI-level regression tests for the chicane/R56 examples

| Gap | Why it matters | Suggested next step |
|---|---|---|
| POSING / pose reference attribute | Right now `X,Y,Z,THETA,PHI,PSI` implicitly mean body pose. Users often think in entrance pose. This ambiguity must be removed. | Add `POSE_REFERENCE = BODY \| ENTRY \| EXIT`, keep internal storage as body pose |
| More placement tests | Current tests cover important pieces, but not enough contracts across all pose interpretations | Add unit tests for body, entry, exit pose conversion once `POSE_REFERENCE` exists |
| More bend regression tests | SBEND examples are useful, but only partially automated | Promote `test-sbend-1`, `test-sbend-2`, and `test-chicane-1` into regression tests with clear tolerances |
| Chicane `R56` validation | Current result is reasonable but not yet a hard regression criterion | Add documented tolerance and maybe a `DT` convergence check |
| OrbitThreader validation | Reporting is improved, but `validateVisitedElements` still needs a sharper definition of expected visited sequence | Test visited element order/count for single bend, SBEND pair, and chicane |
| Pole-face coverage | Edge angles are tested, but we should make sure the matrix of bend angles and pole faces is complete and documented | Expand/lock the 4-angle x 4-pole-face table as executable tests |
| RBEND parity | Most focus has been SBEND | Add equivalent RBEND placement/sign/port tests |
| Manual terminology | Manual now describes the body-pose convention, but POSING terminology is not yet formalized | Add a small "Pose reference convention" subsection |
| Field maps | Explicitly out of scope for this branch | Keep for later branch after analytic bend placement is closed |


aaadelmann:
cscs-ci run cscs-ci-gh200, cscs-ci-mi300

src/Elements/OpalBeamline.cpp
```diff
@@ -21,21 +21,50 @@
...
 #include <fstream>
+#include <iomanip>
 #include <regex>
+#include <sstream>
+#include <vector>
+
+namespace {
+    double normalizeSignedRadians(double angle) {
+        const double pi    = std::acos(-1.0);
+        const double twoPi = 2.0 * pi;
```
pecalvop: It would be better to use `Physics::two_pi`

src/Elements/OpalBeamline.cpp
```diff
@@ -21,21 +21,50 @@
...
+        angle              = std::fmod(angle + pi, twoPi);
+        if (angle < 0.0) {
+            angle += twoPi;
+        }
+        return angle - pi;
+    }
+
+    double radiansToDegrees(double angle) { return angle * 180.0 / std::acos(-1.0); }
```
pecalvop: Why not use `Units::rad2deg`?

rammann:
Ah there is no solver part, my bad! I was confusing it with `DistributionMoments`.

aaadelmann:
Check "Testing" in the above PR  description.

rammann (commented):
Since this is one part of the multi-PR effort to implement 3D element placement in OPALX, I'll also raise some broader points that go beyond this specific PR but are important for the overall structure of the code.

rammann:
#### Element Placement 
The basic pipeline from input file to placed element it the following:

1. Specification in the inputfile using either 
	- ELEMEDGE
	- ORIGIN/ ORIENTATION
	- X/Y/Z/THETA/PHI/PSI
	and the parser accumulates the pathlength along the elements (still called z but should be s)
2. Attribute parsing to the element
	- Calculation of the Lab $\rightarrow$ local element coordinate transformation with the `CoordinateSystemTrafo` class
3. Compile the legacy placement $\rightarrow$ create full 3d placement for elements specified using `ELEMEDGE`.

I think it makes sense to make this the only possible way to specify orientation of elements in space in OPALX. We have a lot of legacy code inside `src/BeamlineGeometry` which is not used. There is alot of overlapping functionality between the `CoordinateSystemTrafo` class and the `3DEuclid`, `Geometry`, `Rotation3D`, and several other classes. In my opinion we should have one fixed way to deal with 3d geometry and placement.

aaadelmann:
This PR will allow us move closer to a V1.0 release because of basic features available. Also from the user point of view all is set: X/Y/Z/THETA/PHI/PSI. 

The we need a few PR's to cleanup and unify some internals, as you pointed out !

rammann:
Sounds good, I'll point out everything I think we should change and then we can decide what we add to this PR and what we put on the todo-list. I'll try to get it done ASAP. 👍

rammann (commented):
### Consistency with regard to registering elements 
So far, every element is mapped into its field-evaluation frame through one uniform call in OpalBeamline: transformToFieldLocalCS() → getFieldCSTrafoLab2Local(comp).transformTo(r), i.e. a single rigid CoordinateSystemTrafo::transformTo.

For `SBEND` and `RBEND` this patterns i broken. Instead we use 
```
if (type == ElementType::SBEND || type == ElementType::RBEND) { 
    const auto* bend = dynamic_cast<const BendBase*>(comp.get());
    if (type == ElementType::RBEND)
        return bend->convertBodyCartesianToFieldLocal(...);         
    return bend->convertEntryCartesianToFieldLocal(...);
}
```
in `OpalBeamline.h`. 
I think it would be good to keep a uniform treatment of the elements. This would mean we have to generalize such that the curvlinear frame necessary for the bends is expressible through the standard interface.

rammann (commented):
### Consistency with applying fields to particles
So far each elements' fields are applied through the `apply(pc)` function. 
```
pc->transformBunch(refToLocalCSTrafo);
(*it)->apply(pc);
pc->transformBunch(localToRefCSTrafo);
```
For `SBEND` and `RBEND` we break the pattern and instead use 
```
if ((*it)->getType() == ElementType::SBEND || (*it)->getType() == ElementType::RBEND) {
    const auto slices = bend->buildTrackingSlices();
    for (const auto& slice : slices) {
        pc->transformBunch(refToSliceLocal);
        bend->applySlice(pc, slice);       
        pc->transformBunch(sliceLocalToRef);
    }
    continue;
}
``` 
in the `ParallelTracker`. 
Here I think it would make sense to move the application of the slices into the `apply(pc)` of the bends instead of explicitly guarding against the element type.

aaadelmann:
Check the last 3 commits they address all consistency points

src/Algorithms/ParallelTracker.cpp
```diff
@@ -745,33 +745,49 @@ void ParallelTracker::computeExternalFields(OrbitThreader& oth) {
...
+                continue;
+            }
+
+            for (const auto& element : segment.getActiveElements()) {
+                if (element->getType() == ElementType::SBEND
+                    || element->getType() == ElementType::RBEND) {
+                    elements.insert(element);
+                }
+            }
+        }
```
rammann: Shouldn't the `elements` array already contain the `SBEND` and `RBEND` after `oth.query(queryCenter, queryHalfWidth)`? What is the second check (lines 762-773) for?

src/Algorithms/ParallelTracker.cpp
```diff
@@ -1273,16 +1289,21 @@ void ParallelTracker::updateReferenceParticles(const BorisPusher& pusher) {
...
 
-        IndexMap::value_t elements           = itsOpalBeamline_m.getElements(pc.getRefPartR());
-        IndexMap::value_t::const_iterator it = elements.begin();
+        IndexMap::value_t elements = itsOpalBeamline_m.getElements(pc.getRefPartR());
+        const auto allElements     = itsOpalBeamline_m.getElements();
+        for (const auto& element : allElements) {
+            if (element->getType() == ElementType::MONITOR && element->Online()) {
+                elements.insert(element);
+            }
+        }
```
rammann: Maybe here we could also unify the registration of `MONITOR` instead of explicitly guarding against it?

rammann:
I left a few more comments but if the regression tests pass I'm ok with approving this. As we said, we can do some internal cleanup in future PRs.

aaadelmann:
Can not see the additional comments

rammann:
I mean the two parts of the code I highlighted above. So `ParallelTracker:757-773` and `ParallelTracker:1294-1298`.

src/Elements/OpalBeamline.cpp
aaadelmann: done

src/Elements/OpalBeamline.cpp
aaadelmann: done

src/Algorithms/ParallelTracker.cpp
aaadelmann: done at least to first order for now

src/Algorithms/ParallelTracker.cpp
aaadelmann: done at least to first order for now

src/Elements/OpalBeamline.cpp
```diff
@@ -21,21 +21,47 @@
...
+#include <algorithm>
 #include <filesystem>
 #include <fstream>
+#include <iomanip>
 #include <regex>
+#include <sstream>
+#include <vector>
+
+namespace {
+    double normalizeSignedRadians(double angle) {
```
pecalvop, line 37: This function is now unused, so it can be cleaned up.

rammann (approved):

src/PartBunch/FieldMirror.hpp
```diff
@@ -130,14 +130,9 @@ namespace opalx {
...
 
-                // recv
-                ippl::Vector<bool, Dim> coordBool = false;
-                if constexpr (Dim > 0) coordBool[0] = flipX;
-                if constexpr (Dim > 1) coordBool[1] = flipY;
-                if constexpr (Dim > 2) coordBool[2] = flipZ;
-
                 ippl::detail::solver_recv(
-                        base_tag, 0, p, intersect, ldom_r, nghost, dstView, fd_recv, coordBool);
+                        base_tag, 0, p, intersect, ldom_r, nghost, dstView, fd_recv, flipX, flipY,
```
pecalvop: This commit causes a compilation error

pecalvop (changes requested):

aaadelmann:
then you are compiling  with the wrong IPPL version please use -DIPPL_GIT_TAG=27d11d0b58bc5db6a582f8132c3d6c88285b26bd 

Cheers A
------
Dr. sc. math. Andreas (Andy) Adelmann
Paul Scherrer Institut OHSA/D17 CH-5232 Villigen PSI
Phone Office: xx41 56 310 42 33 Fax: xx41 56 310 31 91
Zoom ID: 470-582-4086 Password: AdA
Zoom Link: https://ethz.zoom.us/j/4705824086?pwd=dFcvT1pMMGY0bHg0dTNncUNZZTJkZz09

-------------------------------------------------------
Friday: ETH HPK G 28   +41 44 633 3076
============================================
The more exotic, the more abstract the knowledge, 
the more profound will be its consequences.
Leon Lederman 
============================================

pecalvop (approved):

PR #429, opened 2026-05-21
Implement proper `.stat` values for empty bunches

aliemen:
closes #428 

This PR should fix the failing `Dist-flattop` test where the number of written steps in the `.stat` files doesn't match because Gwendolen has a slight numerical differences in particle emission calculation. For example, the first emitting timestep on my Mac for `Dist-opalflattop` would be step 1 with ~2 particles, while Gwendolen would start with them on step 2 (or something similar to that).


aliemen:
Unit tests and regression tests all pass on CPU and GPU (GH200) single and multirank (tested with 1 and 2 ranks locally and on Daint).

aaadelmann (approved):

PR #431, opened 2026-05-22
Update FieldMirror for changed IPPL FieldBufferOps API

MichZampetakis:
Closes https://github.com/OPALX-project/OPALX/issues/430

## Summary

Updates `FieldMirror.hpp` to match the current IPPL `FieldBufferOps.hpp` API.

## Changes

- Use `view_type` as the second template parameter for `solver_send_field`
- Replace separate `flipX`, `flipY`, `flipZ` arguments to `solver_recv` with a single `ippl::Vector<bool, Dim>`

### DIFF
--- a/src/PartBunch/FieldMirror.hpp
+++ b/src/PartBunch/FieldMirror.hpp
@@ -115,7 +115,7 @@ namespace opalx {
                 // Pack r's src data at `intersect` (r's src global coords) and send to p.
                 // p will unpack the buffer with the mirror-axis flip flag, placing
                 // the data at the reflected dst range.
-                ippl::detail::solver_send_field<value_type, value_type, Dim>(
+                ippl::detail::solver_send_field<value_type, view_type, Dim>(
                         base_tag, 0, p, intersect, ldom_r, nghost, srcView, fd_send, requests);
             }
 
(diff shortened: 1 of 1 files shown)


src/PartBunch/FieldMirror.hpp
```diff
@@ -115,7 +115,7 @@ namespace opalx {
                 // Pack r's src data at `intersect` (r's src global coords) and send to p.
                 // p will unpack the buffer with the mirror-axis flip flag, placing
                 // the data at the reflected dst range.
-                ippl::detail::solver_send_field<value_type, value_type, Dim>(
+                ippl::detail::solver_send_field<value_type, view_type, Dim>(
```
aliemen: Perhaps just remove the templates alltogether. Let the compiler deduce them itself, so something like `ippl::detail::solver_send_field(...` should be enough.

src/PartBunch/FieldMirror.hpp
```diff
@@ -130,9 +130,14 @@ namespace opalx {
...
-                        base_tag, 0, p, intersect, ldom_r, nghost, dstView, fd_recv, flipX, flipY,
-                        flipZ);
+                // recv
+                ippl::Vector<bool, Dim> coordBool = false;
+                if constexpr (Dim > 0) coordBool[0] = flipX;
+                if constexpr (Dim > 1) coordBool[1] = flipY;
+                if constexpr (Dim > 2) coordBool[2] = flipZ;
+
+                ippl::detail::solver_recv<value_type, view_type, Dim>(
+                        base_tag, 0, p, intersect, ldom_r, nghost, dstView, fd_recv, coordBool);
```
aliemen: Same here with the templates.

~Also, OPALX is always 3D, so I guess you can just pass `ippl::Vector<bool, Dim>{flipX, flipY, flipZ}` in the argument.~ OK ignore this one, doesn't matter in this case

aliemen (changes requested):
I would just remove the templates alltogether. Makes the code much simpler and future proof should the API change again.

aliemen:
Looks good, thanks for fixing. I'm starting a 2 rank clean build regression run on Daint. If it passes, I'll accept the PR.

For reference (just so the IPPL PR is linked): IPPL-framework/IPPL#520 changed the templates in `solver_send_[...]` to allow e.g. `Dim != 3`. However, this resulted in wrong template deduction from the OPALX side. By removing the passed templates alltogether, the compiler can figure everything out correctly. That's also why todays [regression tests failed](https://amas.pages.psi.ch/opal-live-doc/opalx-regression-test/overview/master/gpu-cuda-a100/index.html#:~:text=2026%2D05%2D22%2001%3A01).

aliemen (approved):
Unit and regression tests passed on CPU and GPU (GH200, 2 ranks).

PR #438, opened 2026-05-29
Test/fix CUDA compilation check

aliemen:
closes #437


aliemen:
OK [seems to work now](https://github.com/OPALX-project/OPALX/actions/runs/26685064809), my guess is that `cuda/13.1` in the runner and the new Kokkos 5 default pulled by IPPL caused the compiler seg-fault in the runner... This kind of issue is not unheard of, see e.g. kokkos/kokkos#8794. 

If we downgrade the runner to `cuda/12.8`, it works again.

aaadelmann:
yea lets get back to 12.8

aaadelmann (approved):

PR #441, opened 2026-06-02
removed duplicate opalparticle

rammann:
Removed duplicate `OpalParticle` class and closes #440.

- Deleted `src/Algorithms/OpalParticle.{h,cpp}`
- Removed the entries from `Algorithms/CMakeLists.txt`

Testing:

- Passing all unit tests and regression tests on CPU and GPU.


rammann:
It seems the clang format is failing because running `clang-format` on a non existent file completes with exit code 1. 
I made and issue #442.

aaadelmann (approved):

PR #445, opened 2026-06-03
CI/CD initial setup for opal-x on cscs alps runners

biddisco:
Ready to merge


biddisco:
cscs-ci run

biddisco:

biddisco:
cscs-ci run cscs-ci-mi300, cscs-ci-gh200, cscs-ci-openmp

CTestConfig.cmake
```diff
@@ -0,0 +1,22 @@
+## This file should be placed in the root directory of your project.
+## Then modify the CMakeLists.txt file in the root directory of your
+## project to incorporate the testing dashboard.
+##
+## # The following are required to submit to the CDash dashboard:
+##   ENABLE_TESTING()
+##   INCLUDE(CTest)
+
+set(CTEST_PROJECT_NAME OpalX)
```
aaadelmann: We call it OPALX

CMakeUserPresets.json
```diff
@@ -0,0 +1,156 @@
...
+        "OPALX_PLATFORMS": "SERIAL;CUDA",
+        "PLATFORMS": "SERIAL;CUDA",
+        "CMAKE_CUDA_ARCHITECTURES": "90",
+        "MPIEXEC_EXECUTABLE": "/usr/bin/srun",
+        "Kokkos_VERSION_DEFAULT": "4.7",
+        "Heffte_VERSION_DEFAULT": "2.4.1"
+      }
+    },
+    {
+      "name": "alps-mi300",
```
aaadelmann, line 85: What is the difference between alps-mi300 and alps-mi300-kokkos?

aaadelmann (changes requested):

CMakeUserPresets.json
biddisco: I've removed the kokkos one. I was experimenting in the past with setting many kokkos vars by hand to force certain configurations

aaadelmann (approved):

aliemen:
_(you probably have this on your mind, but let me mention it anyways)_ I clicked through the opalx cdash and it looks like it's still not configured with the OPALX `cmake`, but through IPPL presets. I don't think all of the IPPL presets propagate through, so I would maybe let OPALX set its IPPL cmake variables internally? 

Also, e.g. [Eiger `openmp`](https://my.cdash.org/builds/3624716/configure) configures `SERIAL` instead of `OPENMP` (I have not checked the rest).

src/Distribution/Distribution.cpp
```diff
@@ -293,9 +293,9 @@ void Distribution::setDistParametersGauss() {
      * on the input attributes.
      * In case of DistributionType::MATCHEDGAUSS we only need to set the cutoff parameters
      */
-
-    cutoffR_m = 3.;
-    cutoffP_m = 3.;
+    emitting_m = true;
```
aliemen: Gaussian distribution is not emitted, so this needs to be `false` (I suppose you added that because of it being potentially uninitialized?)

aliemen (changes requested):

src/Distribution/Distribution.cpp
aliemen: (even better would be to set it like [this](https://github.com/OPALX-project/OPALX/blob/678e76ab4c374c806640a3cb69ffa8328d44fd8b/src/Distribution/Distribution.cpp#L400) and then throw when it's `true`, but it's not that important...)

biddisco: I did not realize I had committed this change. It must have been a compilation fix in my dir that I accidentally committed when adddinf the CI/CD changes. 

Should I remove it (and we do a separeate PR for it), leave it (with a fix as suggested by you), or something else

aliemen: I think removing is fine. 

(If I remember correctly, we talked about this some time ago as a potential fix for the solver seg-faulting - but at this time I'm pretty sure that the segfault has nothing to do with that `emitting_m` flag)

biddisco:
I have removed the unexpected changes to src/ files and add one last commit to show the URL for the dashboard (as done in IPPL). If this passes CI/CD, then it should be ready to merge.

Edit. I will check the OpenMP platforms etc as I think I forgot to fix that

aliemen (approved):
Looks good to me! I've added the cdash badge (similar to IPPL) to our CI section in the README, hope that's OK. 

I did not thoroughly check the files (only glanced over them) in the CI, but I guess since it seems to work it's fine. 

Regression/unit tests all passed on CPU (multirank) and on GH200 (single rank).

PR #449, opened 2026-06-19
Added cmake symbol that can switch off the -ffile-prefix-map option t…

Jonathan-A-Thompson:
…hus allowing the source tree to be referenced by the object files.


Jonathan-A-Thompson:
With this change I can define KEEP_ABSOLUTE_PATHS_FOR_DEBUGGER in my inject.cmake file and that allows the Clion debugger and the Clang coverage tool to find the source files correctly.

aaadelmann (approved):

PR #454, opened 2026-06-23
Remove Legacy Fields 

rammann:
This PR removes the legacy `EMField` classes and the `getField()` accessors. Since the modern elements calculate their fields directly (`apply()`) and store necessary information inside `Kokkos::View`s, these old field classes are unused. 

## Changes

**Deleted the `Fields/` classes:** `AcceleratingField`, `BDipoleField`,
`BMultipoleField`, `ConstBField`, `ConstBzField`, `ConstEzField`, `EMField`,
`NullField`, `OscillatingField`, `StaticElectricField`, `StaticMagneticField`,
plus their CMake entries and `TestConstEzField`.

## BendBase
`BendBase` was the only remaining element that actively used the legacy field classes. It kept its fieldstrength inside a `BMultipoleField` object and during each `apply()` it would then create temporary views, deep-copy to device, and then apply using these views. Now it was changed to directly store the values on a device View. This mirrors the way `Multipole` works and avoids unnecessary deep copies. 

## Testing
Passing all unit and regression tests on CPU and GPU. 

Closes #453.


src/AbsBeamline/BendBase.h
```diff
@@ -156,8 +161,13 @@ class BendBase : public Component {
     double getStoredExitAngle() const;
 
 private:
-    static void computeFieldHost(
-            const Vector_t<double, 3>& R, const BMultipoleField& field, Vector_t<double, 3>& B);
+    void computeFieldHost(const Vector_t<double, 3>& R, Vector_t<double, 3>& B) const;
```
aaadelmann, line 164: add a comment maybe also indicating what is different w.r.t. OPAL

src/AbsBeamline/Corrector.h
```diff
@@ -24,7 +24,6 @@
 #include "AbsBeamline/BeamlineVisitor.h"
 #include "AbsBeamline/Component.h"
 #include "BeamlineGeometry/StraightGeometry.h"
-#include "Fields/BDipoleField.h"
 
 // Class Corrector
```
aaadelmann, line 28: And what if the corrector has a field? This could be mentioned in a comment

aaadelmann:
Add to this PR a class diagram before and after, showing what is removed.

aaadelmann:
I see the apply() and  temporary views issue. However we should be able to keep the abstraction and just pull out the date from the "Fields" outside of apply()

rammann:
class-diagram

aaadelmann (approved):

PR #460, opened 2026-06-24
Calculate Initial Reference Position from Particles

rammann:
This PR addresses the issue #459. 

The initial `RefPartR` is now calculated in the following way:
1. If there are particles at $t=0$, `getMeanR()` of the particles.
2. If there are no particles but a sampler with an initial reference position, `getInitialReferencePosition()`
3. If there are no particles and no sampler, `RefPartR = {0,0,0}`

This mirrors what is already done for `RefPartP`.

closes #459.

### DIFF
--- a/src/Algorithms/ParallelTracker.cpp
+++ b/src/Algorithms/ParallelTracker.cpp
@@ -206,35 +206,59 @@ void ParallelTracker::execute() {
                     "Particle container has null PartData reference during lab-frame init.");
         }
         pc->setToLabTrafo(beamlineToLab);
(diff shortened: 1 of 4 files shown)


rammann:
Dist-fromFile regression test is failing, looking into it!

rammann:
Ok, reason is extremely low tolerances for that test. Since RefPartR is not very slightly different $O(10^{-9})$ the test is failing. Will make a new reference and a PR to the regression tests repo too.

aaadelmann (approved):

PR #471, opened 2026-06-26
Add GREENSF selection option to OPALX

aaadelmann:
Forward the existing OPALX FIELDSOLVER GREENSF option into the IPPL open-boundary FFT solver configuration.

The OPALX parser already exposes GREENSF with the legacy-compatible values STANDARD and INTEGRATED, defaulting to INTEGRATED. This change adds the missing runtime plumbing so the parsed value is actually used when initializing the OPEN field solver:

- add FieldSolverCmd::getGreensFunction()
- store the selected Green-function mode on FieldSolver
- pass GREENSF through PartBunch -> BinnedFieldSolver -> FieldSolver
- map STANDARD/INTEGRATED to IPPL's greens_function parameter in initOpenSolver()
- throw a clear OpalException for unsupported GREENSF values
- add BinnedFieldSolver smoke-test coverage for both explicit modes
- inputfile syntax compatible with OPAL
- 
The shifted Green's function path remains a separate OPALX feature. Once the base OPEN solver is initialized with GREENSF=INTEGRATED, IPPL's shifted Green-function hook will use the corresponding integrated shifted kernel.

Validated with:

- cmake --build build_serial --target TestBinnedFieldSolver -j 8
- ctest --test-dir build_serial -R '^TestBinnedFieldSolver$' --output-on-failure
- mpirun -np 2 build_serial/unit_tests/PartBunch/TestBinnedFieldSolver
- mpirun -np 4 build_serial/unit_tests/PartBunch/TestBinnedFieldSolver
- git diff --check


aliemen:
This looks good on the OPALX side. I merged `add-integrated-greens-function` locally into my `455-...` branch for testing. I am not sure if it is entirely correct. When running SwissFEL again ($32^3$ and 8ppc), I get the following plot.

comparison-integrated

The green and blue lines are OPAL and OPALX with the `STANDARD` Greens function - they match as expected. Then, the orange and red line use the new `INTEGRATED` version. In longitudinal directional it looks much better. However, the new IPPL implementation of the integrated Green's function somehow messes up the transversal direction.

But in any case, this PR can be merged. The issue has to be somewhere on the IPPL side.

aliemen (approved):
Looks good to me on the OPALX side. I am not sure if the IPPL implementation is correct, see my comment above. 

I would say, let's merge this as is. Later, I would perhaps try not to route the `INTEGRATED`/`STANDARD` flags through the constructor in `FieldSolver.cpp`, since it would be a solver-specific flag even though `FieldSolver` is a general class for all solvers. But since I am planning to refactor the solver environment anyways, it's not necessary yet to worry about it.

aliemen:
cscs-ci run cscs-ci-gh200

aliemen:
Alright, I'm just going to merge it. `cscs-ci-gh200` failing seems to be a pipeline error (and I cannot trigger it again because of missing permissions). I ran all regression tests and unit tests on Daint without any issues myself.

PR #476, opened 2026-06-29
Reproduce `SwissFEL-booster`: emission timings/mesh stretching, cathode particle loss and `EKIN`

aliemen:
This PR introduces a variety of adaptations in OPALX that are made in order to exactly match solver related behaviour/code paths from OPAL. I validated everything against old OPAL's `SwissFEL-booster` test, see [my comment](https://github.com/OPALX-project/regression-tests-x/issues/38#issuecomment-4800235287) on OPALX-project/regression-tests-x#38. 

## Changes made

- Emission-step `DT` timing: `ParallelTracker::execute()` now calls `selectDT()` immediately after emission.
- Backward source-plane losses: Kept the source-plane check in `computeExternalFields()`, matching the old OPAL timing. The loss plane is shifted to the legacy `SOURCE` entrance plane, `R0Z - 0.05m`, instead of deleting at the cathode/image plane.
- Early space-charge gating: `computeSpaceChargeFields()` now skips SC until the total active particle count exceeds `MINBINEMITTED` (default `10` - same as OPAL).
- Old OPAL mesh geometry during emission: Field mesh spacing now uses `(upper - lower) / (N - 1)` with origin `lower - 0.5*h`. During emission, the longitudinal mesh is stretched using the emitted fraction, matching old OPAL’s emitting-beam `boundp()` behavior. I introduced a "`emissionMeshStretchActive`" flag that indicates if we're in the beginning of an emission (where the bunch looks like a "pancake"). This flag will be properly integrated with the `BunchStateHandler`. 
- Emission kinetic-energy handling: Added `EKIN` to `EMISSIONSOURCE`. For generated emitted distributions, `EKIN` is converted to `beta*gamma` and applied like old OPAL’s cathode emission momentum. 
- `EMITTEDFROMFILE` replay timing: Parses the optional old OPAL `Bin Number` column. Reconstructs the old emission window and global time shift from bin data or `FlatTop` timing attributes.
- Moment cache after coordinate transforms: `ParticleContainer::transformBunch()` now marks moments dirty. This fixes stale `.stat` values like `mean_s` after reference-frame updates and `DT` switches. This might break some regression test references.
- Shifted Green’s image-charge distance: OPALX shifted-GF now uses the full old-OPAL-equivalent distance: `shift_z = 2 * gamma * (z_center - z_plane)` in IPPL convention. This change was unnoticed before, since it "cancelled out" the mistake from the previous emission timings...

I also introduced a clear separation between `P0X/Y/Z`, `EKIN` and `BEAM_P0`. One-shot distributions get `P = BEAM_P0 + P0X/Y/Z`. Emitted distributions ignore `BEAM_P0` (throw when passed), use `EKIN` to scale sampled momenta (for `ASTRA` it is the "magnitude") and then add `P0X/Y/Z`. That means we have something like (of course imagine `EKIN` was converted to `beta*gamma` units) for `ASTRA`:
```cpp
...
p[0] = P0[0] + EKIN * Kokkos::sin(phi) * Kokkos::cos(angle);
```
For `EMISSIONMODEL=NONE`, `EKIN` is simply added to `P0Z`. This matches exactly old OPAL behavior. 

## Testing

Note that since there are small timing related changes in the emission process, changes in Dirichlet BC definitions and backwards moving particle deletion as well as the shifted Green's path, the following regression tests need to be re-done:
- `AWAGun-1-emittedfromfile`
- `Dist-emittedfromfile`
- `Drift-3-periodic-fromfile`
- `Drift-4-multi-emit-open`
- `Drift-4-open-bins-fromfile`
- `Drift-4m-open-fromfile`

Once this PR is approved, I'll redo all of the references for these tests and add a minimal `SwissFEL-boost`-like test. Unit tests are all passing on CPU as well as on GPU (Daint). The other regression tests also pass just fine (single and multirank on CPU and GPU).

Apart from that I tested `SwissFEL-booster` also with `EMISSIONMODEL=ASTRA`, which before resulted in very weird results (see e.g. [this comment](https://github.com/OPALX-project/regression-tests-x/issues/38#issuecomment-4716021231)). Now it looks good, similar to `EMISSIONMODEL=NONE`:

comparison-astra

Apart from that I also re-ran `AWAGun` just to see if it now looks any better. The transversal values are unchanged, but additionally also the longitudinal coordinates match much better compared to [the old plot](https://github.com/OPALX-project/OPALX/pull/423#issuecomment-4466244671):

comparison-awagun


src/Algorithms/ParallelTracker.cpp
```diff
@@ -719,7 +730,26 @@ void ParallelTracker::computeSpaceChargeFields(unsigned long long step) {
             itsBunch_m->getParticleContainer()->R.getView(),
             itsBunch_m->getParticleContainer()->getLocalNum());
     m << level4 << "Transform particle positions to beam coordinate system done." << endl;
+
+    // Enable the transient old-OPAL emission mesh stretch only for this beam-frame bunchUpdate.
```
aaadelmann: Describe a bit more in detail what is going on here

src/Algorithms/ParallelTracker.cpp
```diff
@@ -760,6 +790,15 @@ void ParallelTracker::computeExternalFields(
     IpplTimings::startTimer(fieldEvaluationTimer_m);
     Inform msg("ParallelTracker ", *gmsg);
 
+    // Source-plane loss has to be handled here, before external fields are evaluated. Deferring it
```
aaadelmann, line 796: Define better what source-plan losses are

src/BasicActions/Option.cpp
```diff
@@ -143,6 +144,12 @@ Option::Option()
             "A value of 0 disables these status lines; its default value is 1.",
             stepInfoFreq);
 
+    itsAttr[PRINTRANKDISTRFQ] = Attributes::makeReal(
```
aaadelmann, line 147: is that steering the lbal file write frequency?

aaadelmann (changes requested):

src/BasicActions/Option.cpp
aliemen: Ah no, sorry I forgot to explain this - I can remove it if it's redundant. The `OPTION` is turned off by default. 

I didn't know we already had a similar mechanism (these `.lbal` files). I added this flag for debugging purposes. [`ParticleContainer::printRankLoadInfo`](https://github.com/OPALX-project/OPALX/blob/8c5248a994de736163c18afe8a0bced1359f8e03/src/PartBunch/ParticleContainer.hpp#L751) is called with this frequency, which prints out the particle count distribution across ranks, e.g.:
```text
ParticleContainer::printRankLoadInfo{0}[2]> Particle load by rank (container[0])
ParticleContainer::printRankLoadInfo{0}[2]> Metric                   |       Rank 0 |       Rank 1 |       Rank 2 |       Rank 3 |       Rank 4 |       Rank 5 |       Rank 6 |       Rank 7
ParticleContainer::printRankLoadInfo{0}[2]> Particles [%]            |         0.35 |        10.88 |        25.74 |        19.48 |        18.91 |        16.85 |         7.62 |         0.18
ParticleContainer::printRankLoadInfo{0}[2]> Allocated memory [%]     |        11.95 |        18.45 |        17.16 |        15.97 |        14.72 |        11.16 |         6.55 |         4.05
```

src/Algorithms/ParallelTracker.cpp
```diff
@@ -443,10 +439,10 @@ void ParallelTracker::execute() {
...
             m << level4 << "Emit particles from emission sources done at step " << step << "."
               << endl;
-            const size_t nSourceMarkedAfterEmission = markBackwardParticlesAtSourcePlane();
-            if (nSourceMarkedAfterEmission > 0) {
-                deleteInvalidParticles(true, m, "backward source-plane particles after emission");
-            }
+            // Old OPAL reselects the global dt after emission. On the final emission step this
+            // switches getdT() back to the track step before external fields, reference update, and
+            // time increment, while per-particle fractional dt values remain untouched.
+            selectDT();
```
MichZampetakis, line 445: Is this the reason why OPAL and OPALX produce `.stat` files of different size?

src/Algorithms/ParallelTracker.cpp
```diff
@@ -945,6 +984,11 @@ size_t ParallelTracker::markBackwardParticlesAtSourcePlane() {
 
     const double sourcePlaneZ = imageChargeConfigured ? bsolver->getImageChargePlaneZ()
                                                       : bsolver->getShiftedGreensPlaneZ();
+    // Legacy OPAL's SOURCE element is 5 cm long and is shifted upstream from ELEMEDGE.
+    // Source::apply deletes only once a particle crosses the element-local entrance plane
+    // (Rz <= 0), not when it crosses the cathode/image plane at ELEMEDGE.
+    constexpr double legacySourceLength = 0.05;
+    const double sourceLossPlaneZ       = sourcePlaneZ - legacySourceLength;
```
MichZampetakis, line 999: As it is written, in OPAL there is the `GunSource` element which defines the loss plane and the image charges. Is this hardcoded now in OPALX, or I understood something wrong?

src/Algorithms/ParallelTracker.cpp
aliemen: No, this shouldn't change any physics. Before the fix, `mean_x/y/s` calculation would use values from the moments that were calculated from the previous timestep. Basically, this fixed the `mean_s` discrepancies we observed.

The reason we have potentially different `.stat` lengths is [this line](https://github.com/OPALX-project/OPALX/blob/8c5248a994de736163c18afe8a0bced1359f8e03/src/Algorithms/ParallelTracker.cpp#L1553). I changed it such that active containers also write `.stat` rows even if they don't have particles. That means OPALX is much more resistent to some "statistical inaccuracies" (like sometimes with low particle count or high resolution `FlatTop`, particles might start to emit in different timesteps).

src/Algorithms/ParallelTracker.cpp
aliemen: Yes, the same functionality is implicitly activated by setting `SHIFTED_GREENS_FUNCTION = true;` and `R0Z` in the `EMISSIONSOURCE`. 

Having the `GunSource` element makes sense, but since it is more of a solver related thing, OPAL separated source plane mechanics from Dirichlet BCs. This creates a lot of ambigious behaviour when activating the one, but not the other... That's why Andy and I decided to not introduce `GunSource` in OPALX, but keep it implicit through `EMISSIONSOURCE` (this was for #423).

src/BasicActions/Option.cpp
aliemen: I just looked it up, the `LBalWriter` object doesn't work yet in OPALX. `DataSink::dumpSDDS()` doesn't call it and `LBalWriter::write()` only has hardcoded dummy values. I would perhaps say, we leave `PRINTRANKDISTRFQ` in for now and I make an issue that `LBalWriter` needs to be wired up (+ `PRINTRANKDISTRFQ` then needs to be removed)?

src/Distribution/OpalFlatTop.cpp
```diff
@@ -357,9 +357,9 @@ void OpalFlatTop::generateLocalParticles(
                     const double rand2 = generator.drand(0.0, 1.0);
                     const double phi   = 2.0 * Kokkos::acos(Kokkos::sqrt(rand1));
                     const double angle = twoPi * rand2;
-                    p[0]               = pTot * Kokkos::sin(phi) * Kokkos::cos(angle);
-                    p[1]               = pTot * Kokkos::sin(phi) * Kokkos::sin(angle);
-                    p[2]               = pTot * Kokkos::fabs(Kokkos::cos(phi));
+                    p[0] += pTot * Kokkos::sin(phi) * Kokkos::cos(angle);
+                    p[1] += pTot * Kokkos::sin(phi) * Kokkos::sin(angle);
```
MichZampetakis, line 361: When I added `+=` in the transverse planes in OPAL, I was getting some field related errors half-way of the simulation. I did not investigate more, but keeping that only in the longitudinal plane solved the issue. Maybe it is worth checking

aliemen: I ran the simulation with `ASTRA` on my branch without any issues (see my plot in the PR description). For proper momentum conservation I would argue that you need it in all directions, otherwise 

```math
\lVert P \rVert = \lVert P_0 + P_\mathrm{tot} \begin{pmatrix} \sin \phi \sin \theta \\ \sin\phi  \sin\theta \\ |\cos \phi| \end{pmatrix}  \rVert \neq \lVert \begin{pmatrix} P_\mathrm{tot} \sin \phi \sin \theta \\ P_\mathrm{tot} \sin\phi  \sin\theta \\ P_0 + P_\mathrm{tot} |\cos \phi| \end{pmatrix}  \rVert .
```

Or do you mean something else?

Maybe you had some initial momentum with `BOUNDPDESTROY` deactivated? Then you would get a huge simulation domain from "transverse scattered particles" which would result in extremely inaccurate fields.

src/Algorithms/ParallelTracker.cpp
aliemen: Done.

src/Algorithms/ParallelTracker.cpp
aliemen: Done.

aliemen:
Note that the `Smoke Test (Drift-4-multi-emit-open)` check will only pass once OPALX-project/regression-tests-x#46 is merged. 

I ran unit and regression tests again (with the new `regression-tests-x` branch linked above). Everything passed single- and multirank on CPU as well as GPU (GH200 on Daint).

src/BasicActions/Option.cpp
aaadelmann: ok then let keep your writer.

aaadelmann (approved):

unit_tests/Distribution/TestOpalFlatTop.cpp
```diff
@@ -208,15 +208,23 @@ TEST_F(OpalFlatTopTest, ProvidesOldOpalInitialReferenceMomentum) {
...
     sampler.setEmissionOffsets(0.0, P0, 0.0, "ASTRA");
+
+    // Since ASTRA is only scaled by emissionMomentumMagnitude, P0 still needs to be additive
     const Vector_t<double, 3> refP = sampler.getInitialReferenceMomentum();
-    EXPECT_DOUBLE_EQ(refP[0], 0.0);
-    EXPECT_DOUBLE_EQ(refP[1], 0.0);
-    EXPECT_DOUBLE_EQ(refP[2], 0.5 * std::sqrt(dot(P0, P0)));
+    EXPECT_DOUBLE_EQ(refP[0], P0[0]);
+    EXPECT_DOUBLE_EQ(refP[1], P0[1]);
+    EXPECT_DOUBLE_EQ(refP[2], P0[2]);  // 0.5 * std::sqrt(dot(P0, P0)));
```
MichZampetakis, line 229: This ASTRA test currently uses the default `emissionMomentumMagnitude = 0.0`, so it only checks that `P0` is preserved. Do you think we should add a nonzero-`EKIN` ASTRA test that actually samples particles and verifies the sampled momenta are scaled by `EKIN` in all components, while `P0X/Y/Z` remains an additive offset, as mentioned in the comment of line 225?

src/Algorithms/ParallelTracker.cpp
MichZampetakis: Ah, I see. Thanks

src/Algorithms/ParallelTracker.cpp
MichZampetakis: Yes, I agree it is better and cleaner like that. Just for me to keep in mind for future reference.

unit_tests/Distribution/TestOpalFlatTop.cpp
aliemen: Yes, I think that's a good idea. I'll add another small test to `ProvidesOldOpalInitialReferenceMomentum` that samples some particles and checks their `P` magnitude with `ASTRA` and non-zero `EKIN`.

src/Distribution/OpalFlatTop.cpp
MichZampetakis: Yes, I agree that applying `P0` as an offset in all components is the physically consistent choice. I only wanted to mention what I observed during some quick OPAL testing: I had an `OFFSETPZ` that was not included when using `EMISSIONTYPE=ASTRA`, and when I tried adding the offset with `+=` in all components I hit a field-related error later in the simulation. I did not investigate it deeply, and when I only applied the offset longitudinally the error disappeared. `BOUNDPDESTROYFP` should have still been at its default value.
Since your OPALX test runs without issues, I would not argue against this change. I agree this is the correct implementation; I just wanted to leave the observation here in case similar field errors appear later.

unit_tests/Distribution/TestOpalFlatTop.cpp
aliemen: Done.

MichZampetakis:
Everything looks fine to me. I also tested your branch again on Gwendolin and I get similar results.
Maybe in the future we should look into improving the initial `.stat` file output, where some values appear as approximately zero, but that seems like a small detail for now.
Thank you, nice work!

src/Distribution/OpalFlatTop.cpp
aliemen: Hm I see. I would guess that maybe there was some interplay with other (now fixed) "bugs" that lead to this field error.

MichZampetakis (approved):

unit_tests/Distribution/TestOpalFlatTop.cpp
MichZampetakis: Thanks!

PR #479, opened 2026-07-01
Activate load balancer in OPALX and unify field storage/layout update

aliemen:
closes #470 

This PR finally activates load balancing in OPALX. I made another PR that needs to be merged first, see IPPL-framework/ippl#560. The corresponding IPPL PR generalizes the ORB class to allow arbitrary domain decompositions - currently it's only possible in all (3) dimensions. 

The second big change is that ALL fields (including scratch fields from the solver) now live inside `FieldContainer`. That makes it way easier to update all layouts at once. After the layout update, the solver has to be re-initialize. This step is necessary in order to re-initialize solver internal scratch fields after the changed layout. 

The third, minor, change is an option that enables the rank index per particle to be printed to H5. That way I can visualize rank distribution in ParaView. 

I ran all of the unit tests and regression tests on CPU as well as on GPU (up to 4 ranks on GH200 and up to 8 ranks on CPU) without any problems. I also tested a few different possible decompositions and visualized them in ParaView, see e.g. [here](https://github.com/IPPL-framework/ippl/pull/560#issue-4785071090). Everything seems to be working as expected. 

## How to use load balancing as a user

The input file now has three switches:
```text
OPTION, REPARTFREQ = 200; // default is "check every 10 steps"
OPTION, LOADBALANCINGTHRESHOLD = 0.2; // default is 0.05 (local particle count is 5% away from perfectly balanced)
...
FS_SC: 	Fieldsolver, 
                ... 
		PARFFTX = false, 
		PARFFTY = false, 
       	PARFFTZ = true, 
		...;
...
```
Load balancing is on by default. It checks the "balance" criterium every `REPARTFREQ` steps (default 10). Then it repartitions in all of the axis that are set to `true` in the Fieldsolver structure. 

## Detailed Changelog

**Load Balancing / ORB**
- Reactivate primary-container ORB load balancing behind `OPTION, REPARTFREQ`.
- Adds `OPTION, LOADBALANCINGTHRESHOLD` with range validation and use it to gate actual repartitioning.
- Removes OPALX first-repartition state; OPALX always repartitions from current particle positions. 
- Delegates decomposition-aware cuts to IPPL via `orb.binaryRepartition(*R, *fl, false, fl->isParallel())`, this needs the new ORB changes in IPPL.

**Field/Layout Refresh**
- Moves self-field scratch ownership into `FieldContainer`: `rho`, `E`, `phi`, `Etmp`, `Btmp`, shifted-Greens flipped-z scratch.
- Adds one central field-layout refresh path after ORB or grid layout changes.
- Updates primary particle layout and migrates particles after successful ORB.
- Adds `FieldSolver::refreshAfterFieldLayoutChange()` to rebind existing solver backends to refreshed fields. This only exists to update solver internal scratch fields.
- Updates `BinnedFieldSolver` to use `FieldContainer` temporary fields and shifted-Greens scratch.

**Diagnostics / Output**
- Adds `OPTION, PRINTRANKDISTRFQ` and `ParticleContainer::printRankLoadInfo()` for per-rank particle/capacity diagnostics. Default it's turned off.
- Adds optional H5 particle-rank dumping behind `OPTION, RANKDUMP=true`; default is off. This can be used to visualize the rank distribution e.g. in ParaView.

And in order to have a nice graph, here is a diagram explaining when and how repartitioning is triggered in OPALX:

```mermaid
flowchart TD
    A["Input OPTION REPARTFREQ = N"] --> B["Options::repartFreq"]
    B --> C["ParallelTracker::setOptionalVariables"]
    C --> D{"N <= 0"}
    D -->|yes| D0["Disable repartition"]
    D -->|no| E{"MPI ranks < 2"}
    E -->|yes| E0["Disable on one rank"]
    E -->|no| F{"Rank count power of two"}
    F -->|no| F0["Warn and disable"]
    F -->|yes| G["repartFreq_m = N"]
    G --> H["Each tracking step"]
    H --> I["computeSpaceChargeFields step"]
    I --> J{"Cadence hit"}
    J -->|no| K["Go to computeSelfFields"]
    J -->|yes| L["LoadBalancer balance"]
    L --> M{"Imbalance above threshold"}
    M -->|no| K
    M -->|yes| N["ParallelTracker doBinaryRepartition"]
    N --> O["PartBunch do_binaryRepart"]
    O --> P["LoadBalancer repartitionFromCurrentParticles"]
    P --> Q["IPPL ORB repartitions with allowedAxes"]
    Q --> R["Field and particle layouts refreshed"]
    R --> K
```


aliemen:
Note: this PR needs IPPL-framework/ippl#560 to be merged first. Otherwise the CI won't pass.

aliemen:
I change the ORB algorithm to only use the "all parallel" version and added documentation in the code how to re-activate the more general version once IPPL-framework/IPPL#559 is merged. 

The current version now **works perfectly fine with IPPL master**, but has one additional constraint: in the input file, you have to set all dimensions to parallel in the `FieldSolver` (since that's the only supported decomposition), otherwise you'll get an exception. Meaning, I need to update all regression test input files to the following structure:
```text
FS_SC: Fieldsolver, 
    ...
    PARFFTX = true,
    PARFFTY = true,
    PARFFTZ = true,
    ...
```
Currently most have `PARFFTX = PARFFTY = false`, which is not supported (yet).

src/PartBunch/FieldContainer.hpp
```diff
@@ -54,6 +55,9 @@ class FieldContainer {
     VField_t<T, Dim> E_m;
     Field_t<Dim> rho_m;
     Field<T, Dim> phi_m;
+    std::shared_ptr<VField_t<T, Dim>> Etmp_m;
+    std::shared_ptr<VField_t<T, Dim>> Btmp_m;
+    std::shared_ptr<VField_t<T, Dim>> flippedZSlabField_m;
```
aaadelmann, line 85: Why do we need flippedZSlabField_m?

aaadelmann (changes requested):

src/PartBunch/FieldContainer.hpp
aliemen: They already existed before, I just moved them into the `FieldContainer` to have all fields in a unified place so none get left behind during the ORB layout update.
- `Etmp_m`: scratch field that's used by `BinnedFieldSolver` to accumulate the `E` field from each bin. 
- `Btmp_m`: similar to `Etmp_m`, it accumulates the `B` field per bin due to the Lorentz transformations.
- `flippedZSlabField_m`: needed for the shifted Green's "field mirror" operation. The "mirrored" rank writes into this field. Since it needs communication, an in-place "swap" would have been too difficult for now (not sure if it is even possible) - so I used another scratch field. Note that I cannot reuse any of the other fields, since all vector fields are in active use during the binning step. The mirror operation needs to be done after the solve and before the accumulate to temp.

aliemen:
All regression and unit tests passed on CPU as well as on GPU (GH200), tested with 2 ranks.

src/PartBunch/FieldContainer.hpp
aaadelmann: Add the description of flippedZSlabField_m to the hpp

aaadelmann (approved):

PR #486, opened 2026-07-09
Add CSCS regression-test pipeline for OpenMP, GH200, and MI300

biddisco:
This tests the regression test execution on CSCS alps machines
No submission of results can happen yet. Awaiting keys to submission repo


biddisco:

aliemen:
Hey, I wanted to see how the regression tests are going and noticed a few cmake things I would like to mention:
- Default build type is`RELEASE`. If you don't use the OPALX options as described [here](https://github.com/OPALX-project/OPALX#building-opalx), your `CMAKE_BUILD_TYPE` will internally always be overwritten with `RELEASE`. So all of the `DEBUG` tests are actually `RELEASE`. 
- The `OPENMP` test is [still `SERIAL`](https://github.com/OPALX-project/OPALX/pull/445#issuecomment-4738346715). Again, OPALX needs to be configured with the [OPALX flags](https://github.com/OPALX-project/OPALX#building-opalx). `OPALX_PLATFORMS` is not supposed to be user facing: it will be overwritten by `PLATFORMS`, which defaults to `SERIAL`.
- Same with `Kokkos_ARCH_FLAG`: OPALX uses `ARCH`, meaning the builds only get the architecture/compute capability by accident from Kokkos' auto detection, not from the cmake.
- I think the `Kokkos_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK` does nothing: it's cleared by OPALX before kokkos configuration.

Overall, everything you should have to set are `ARCH`, `PLATFORMS` and `BUILD_TYPE`. If you want to pin them, the usual `Kokkos_VERSION` and `Heffte_VERSION` are fine. HIP is a bit different, here we don't (unlike CUDA) auto derive `CMAKE_HIP_ARCHITECTURES`, so you have to add it, i.e.:
```bash
cmake .. -DBUILD_TYPE=Release -DPLATFORMS=HIP -DARCH=AMD_GFX942_APU -DCMAKE_HIP_ARCHITECTURES=gfx942
```
Then you should stop passing `DKokkos_ARCH_FLAG`. _I think_ if you do all that, `AMDGPU_TARGETS`, `GPU_TARGETS` and `Heffte_ENABLE_ROCM` are then unnecessary (but I could be wrong with that last point, since I don't have access to any AMD cluster).

---

(the idea with OPALX was that we don't want to overcomplicate things for the user - all flags a user should care about are `ARCH` and `PLATFORMS`)

src/CMakeLists.txt
```diff
@@ -90,8 +90,7 @@ target_include_directories(
   PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 
 
-target_sources(opalx PUBLIC changes.cpp)
-#
```
aaadelmann, line 93: why is changes.cpp removed

ci/cscs/regression-test-mpirun-wrapper
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
aaadelmann, line 1: why do we need this?

aaadelmann (approved):

src/CMakeLists.txt
biddisco: The file was added twice. So I removed one copy. (line 58 I think)

ci/cscs/regression-test-mpirun-wrapper
biddisco: That was so that the existing script could call mpi when we use slurm srun on our system - a shim script to get stuff working without making changes to the original wherever possible

aaadelmann (approved):

PR #487, opened 2026-07-10
Revert temporary fixes that circumvented the RELEASE mode CUDA seg-faults

aliemen:
closes #426 

This PR removes all of the helper that were added to OPALX in order to circumvent the bugs described in the attached issue. These bugs should all be gone after IPPL-framework/ippl#561 and IPPL-framework/ippl#564, meaning these two IPPL PR's need to be merged first.

After this PR, `accumulateHalo` uses its correct IPPL implementation again and the field operations use the implicit assignment operators.


aliemen:
As mentioned in the linked IPPL PR, everything works perfectly fine when built with `-DIPPL_GIT_TAG=investigate-accumulate-halo-segfault`. Regression tests pass multirank on GH200 in Release mode without any of the previously added helpers.

aliemen:
@aaadelmann since the two IPPL fixes are merged, we can also safely merge this one. 

Regression/unit tests all pass on CPU and GH200 (tested with two ranks in release!), including all solver tests.

aaadelmann (approved):

PR #491, opened 2026-07-15
Add missing beam rigidity scaling term in multipole expansion

rammann:
This PR addresses the issue #490 by adding the correct beam-rigidity scaling term to the normal and skew components.

Closes #490.


rammann:
Naturally, the regression tests involving multipoles/quadrupoles will fail now.
Will make a PR to the regression tests repo.

aaadelmann (approved):

PR #496, opened 2026-07-28
fix-dump-frequ

aaadelmann:
if dump frequency is zero no output is generated

### DIFF
--- a/src/BasicActions/Option.cpp
+++ b/src/BasicActions/Option.cpp
@@ -500,16 +500,14 @@ void Option::execute() {
     if (itsAttr[PSDUMPFREQ]) {
         psDumpFreq = int(Attributes::getReal(itsAttr[PSDUMPFREQ]));
         if (psDumpFreq < 0) {
-            throw OpalException("Option::update",
-                                "PSDUMPFREQ must be non-negative.");
+            throw OpalException("Option::execute", "PSDUMPFREQ must be non-negative.");
         }
     }
 
     if (itsAttr[STATDUMPFREQ]) {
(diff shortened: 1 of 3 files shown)


src/BasicActions/Option.cpp
```diff
@@ -498,12 +499,18 @@ void Option::execute() {
 
     if (itsAttr[PSDUMPFREQ]) {
         psDumpFreq = int(Attributes::getReal(itsAttr[PSDUMPFREQ]));
-        if (psDumpFreq == 0) psDumpFreq = std::numeric_limits<int>::max();
+        if (psDumpFreq < 0) {
+            throw OpalException("Option::update",
+                                "PSDUMPFREQ must be non-negative.");
```
aliemen: I think this should be `"Option::execute"`, not `"Option::update"`.

src/BasicActions/Option.cpp
```diff
@@ -498,12 +499,18 @@ void Option::execute() {
...
+        }
     }
 
     if (itsAttr[STATDUMPFREQ]) {
         statDumpFreq = int(Attributes::getReal(itsAttr[STATDUMPFREQ]));
-        if (statDumpFreq == 0) statDumpFreq = std::numeric_limits<int>::max();
+        if (statDumpFreq < 0) {
+            throw OpalException("Option::update",
```
aliemen: Same here

aliemen (changes requested):
Nice! One thing though: at the end of the tracker loop we still print "`Dump phase space of last step`". Perhaps we should guard that too?

aliemen (approved):
OK I applied the two small fixes myself and clang-formatted. lgtm

PR #512, opened 2026-08-06
Add guards against wrong indices for `TestSolve2d5`

aliemen:
closes #511 

See attached issue for an explanation. 

@Jonathan-A-Thompson: it would be nice if you could quickly say if the fix makes sense as is.


rammann (approved):
lgtm

PR #514, opened 2026-08-06
Make adaptbins backend aware

aaadelmann:
**Title**

Make AdaptBins team histogram launch backend-aware

**Description**

This PR makes the team-based local histogram reduction in `AdaptBins` safer for Kokkos 5.x backend limits.

The previous implementation hardcoded:

```cpp
team_size = 128
block_size = team_size * 8
```

and launched without checking whether the active Kokkos backend accepted that team size or the requested per-team scratch memory.

The updated implementation keeps `128` as the preferred team size, but asks Kokkos for the backend’s maximum supported team size before launch. If `128` is too large, OPALX now reports the adjustment and launches with the backend-supported value. The particle block size remains fixed at the previous `128 * 8`, so the particle work assigned per team is unchanged.

The code also checks the requested team scratch memory against `scratch_size_max(0)` before launching. If the per-team histogram scratch memory does not fit, OPALX raises a clear `OpalException` instead of letting Kokkos abort from inside the backend.

**Validation**

- `cmake --build build_kokkos520 -j 8`
- `ctest --test-dir build_kokkos520 --output-on-failure`

Result:

```text
100% tests passed, 0 tests failed out of 76
```


aliemen (changes requested):
Looks good to me. Everything compiles and runs (unit tests, regression tests on CPU and multirank GPU). 

However, I would make it simpler. I remember I didn't use `Kokkos::AUTO`, because I would need the value before the kernel launch for indexing and a kernel launch in order to retrieve the team size. I didn't want two kernel launches and when I wrote it, I didn't know about `sizing_policy.team_size_max`... So I chose 128 as a "never leads to problems on modern GPUs" solution. 

That also means, the clamping to `std::min(preferred, Kokkos::AUTO)` is unnecessary and we can just extract and then use `Kokkos::AUTO` directly. I would even go a step further: I chose "8 particles per thread" in order to minimize atomics in the team thread range for realistic binning scenarios (~10 bins). So we can easily just keep that fixed, use `Kokkos::AUTO`, keep your scratch size check, and calculate indices based on that. 

I will make a quick commit so you see what I mean - feel free to revert should you disagree.

aliemen (approved):

PR #517, opened 2026-08-10
Aperture Implementation Part 2 - Basic Aperture Functionality

rammann:
# Overview
This PR implements basic **z-axis symmetric** aperture functionality for all elements. 

- The `ElementBase::aperture_m` member variable is a `std::pair<ApertureType, std::vector<double>>` uniquely specifies the elements aperture shape. 
- Testing if a position is inside the aperture is delegated to the stateless namespace `ApertureHelper`, which defines `isInsideAperture` as a `KOKKOS_INLINE_FUNCTION` such that it can be called from inside a kernel. 
- Application of the aperture to the particles is performed in the `ElementBase::markOutsideAperture(pc)` function, which is called each iteration of the `ParallelTracker::execute()` loop.

```mermaid
classDiagram
    direction LR

    class ApertureType {
        <<enumeration>>
        RECTANGULAR
        ELLIPTICAL
    }

    class ApertureHelper {
        <<namespace, stateless>>
        +isInsideAperture(x, y, type, xLimit, yLimit)
        +isInsideAperture(r, aperture)
    }

    class ElementBase {
        #aperture_m
        #deleteOnTransverseExit_m
        +setAperture(type, args)
        +getAperture()
        +isInside(r)
        +markOutsideAperture(pc)
        +getGeometry()
    }

    class SBend {
        +markOutsideAperture(pc)
    }


    ElementBase <|-- SBend


    ElementBase *-- ApertureType : aperture_m.first
    ElementBase ..> ApertureHelper : isInside(r)
```
The runtime logic is presented in the diagram below. The application of the aperture follows the same pattern as the application of the $EM$ fields. Particles outside the aperture are marked and deleted at the end of each iteration. 
```mermaid
sequenceDiagram
    participant Deck as input .in
    participant P as OpalParser
    participant OE as OpalElement
    participant EB as ElementBase / SBend
    participant PT as ParallelTracker
    participant PC as ParticleContainer

    Deck->>P: D1: DRIFT, L=1, APERTURE="CIRCLE(0.04)"
    P->>OE: parse() stores raw string in itsAttr[APERT]
    P->>OE: update()
    OE->>OE: getApert() → ELLIPTICAL, 0.02 x 0.02
    OE->>EB: setAperture(type, halfWidths)
    Note over EB: overwrites the 1e6 ctor default

    loop each time step
        PT->>PT: applyElementApertures()
        PT->>PC: transformBunch(ref → element local)
        PT->>EB: markOutsideAperture(pc)
        EB->>EB: skip if deleteOnTransverseExit_m is false
        EB->>PC: Kokkos reduce, InvalidMask |= outside
        PT->>PC: transformBunch(local → ref)
    end
    PT->>PC: end-of-step deletion flushes InvalidMask
```
# Regression and Unit Tests
All unit and regression tests are passing on GPU and CPU

# Validation 
To test I used a simple setup consisting of a drift with a given aperture (either ellipse or rectangle) and a monitor before and after the drift. The monitor outputs before and after the element are given on the left and right, respectively.
image

image

From the .stat file we can see that the particle are deleted correctly at the position of the element (0.5m).
image

In Paraview:
image
image

# Future Work
In a future step we can implement saving which particles are lost where to a file.


aaadelmann (approved):
Make a comment that the lost particles are not yet saved. 
(also not important for now)

When the failing unit tests are fixed you can merge.

PR #524, opened 2026-08-12
Add checkpoint and restart support

### LINKED ISSUE #523: Add Checkpoint/Restart capability
### Title

Add Checkpoint/Restart capability 

### Description

OPALX should support periodic HDF5 checkpoints and continuation of an interrupted simulation from a selected checkpoint.

The checkpoint frequency is controlled through the global option:

```opal
OPTION, CHECKPOINTFREQ=N;
```

`CHECKPOINTFREQ=0` is the default and disables checkpoint creation. A positive value writes a checkpoint every `N` integration steps.

A simulation is restarted from the command line:

```bash
opalx --restart checkpointFile.h5 input.in
```

### Required behaviour

A checkpoint must contain enough state to continue tracking without resampling or reinitializing the beam, including:

- Global simulation step, time and time step
- Tracking-schedule position
- Particle coordinates, momenta, IDs, charge and mass
- Per-container reference-particle and path state
- Active-container state
- Optional spin data
- Autophased cavity phases

Restart checkpoints should be independent of the MPI decomposition. A checkpoint written with one number of MPI ranks should be restartable with another number of ranks.

### Output continuation

A restart is an append run. All existing output should continue under its original filename.

This includes at least:

- Statistics files (`.stat`)
- Load-balancing files (`.lbal`)
- Phase-space HDF5 files (`.h5`)
- Monitor, loss, design-path and other append-aware output

Because an output file may contain records written after the selected checkpoint, OPALX must first rewind it to the checkpoint position and then append the restarted trajectory. Repeated restarts from the same checkpoint must not produce duplicated rows or HDF5 steps.

A process-wide restart state and append mode should be available so that every output-producing component can distinguish a restart from a normal run.

### Additional requirements

- Checkpoint writing must work when regular phase-space HDF5 output is disabled.
- Missing output files may be recreated for the restarted part of the simulation.
- Same-rank uninterrupted and restarted runs should reproduce identical output.
- Restarts using a different MPI rank count should be numerically equivalent; small last-bit differences from MPI reduction ordering are acceptable.

### Initial limitations

Restarting time-dependent or delayed emission and stochastic processes such as particle decay requires preservation of sampler and random-number-generator state. These cases may initially be rejected with a clear error message.

aaadelmann:
## PR title

Add checkpoint and restart support

### Summary

This PR adds application-level checkpoint/restart support to OPALX. Simulations can periodically save their complete tracking state and later continue from a selected checkpoint, including when the restart uses a different number of MPI ranks.

Restarted simulations continue writing to their original output files. Output produced after the selected checkpoint is removed before the continuation is appended, preventing duplicate statistics rows or HDF5 steps.

### User interface

Periodic checkpointing is controlled by the global option:

```opal
OPTION, CHECKPOINTFREQ=N;
```

- `CHECKPOINTFREQ=0` is the default and disables checkpoint creation.
- A positive value writes a checkpoint after every `N` completed integration steps.
- Checkpoints are stored as `<input-basename>_checkpoint.h5`.

A checkpoint is restored with:

```bash
opalx --restart checkpointFile.h5 input.in
```

### State preserved in a checkpoint

The version-1 checkpoint format stores the state needed to continue tracking without regenerating the initial distribution:

- Global integration step, simulation time and time step
- Tracking-schedule segment and offset
- Active and stopped container state
- Per-container path position and reference-particle state
- Reference-frame coordinates
- Particle positions and momenta
- Particle IDs, bins and individual time steps
- Charge and mass in both supported storage modes
- Optional spin coordinates
- Autophased cavity phases

The input file still defines the accelerator and beam configuration. During restart, OPALX validates that its beam containers, species, charge/mass representation, spin configuration and tracking schedule are compatible with the checkpoint.

### MPI portability

Checkpoint particle data is stored globally rather than using the original rank-local decomposition. During restart, particles are redistributed across the currently selected MPI ranks.

This allows, for example:

```text
1-rank checkpoint → 2-rank restart
2-rank checkpoint → 1-rank restart
```

A restart using the same MPI decomposition reproduces the uninterrupted output exactly. Changing the decomposition can introduce last-bit floating-point differences because MPI reductions occur in a different order. Cross-rank validation in this PR used an absolute HDF5 comparison tolerance of `1e-12`.

### Output continuation

Restart is represented globally through `OpalData::inRestartRun()` and selects the process-wide `OpalData::OpenMode::APPEND`.

Existing diagnostic files retain their original names; no `_restart` files are created.

Before appending:

- `.stat` files are rewound to the saved path position.
- `.lbal` files have the corresponding later rows removed.
- Phase-space `.h5` files are rewound to the saved simulation time.
- Monitor, loss, design-path and other append-aware writers inherit the global append mode.

This handles checkpoints that are older than the latest existing diagnostic output. Repeating a restart from the same checkpoint does not duplicate rows or HDF5 steps.

If an expected output file is missing, OPALX recreates it for the restarted portion of the simulation.

Checkpoint creation is independent of regular phase-space output and therefore also works with:

```opal
OPTION, ENABLEHDF5=FALSE;
```

### Current limitations

Checkpoint format version 1 does not preserve emission-sampler or stochastic-process random-number-generator state.

To avoid silently producing a non-reproducible continuation, restart currently rejects:

- Delayed or time-dependent emission
- File-based emission
- Stochastic global processes such as `DECAY`

Iterative field-solver history is reconstructed rather than serialized. Consequently, changing the MPI decomposition is expected to be numerically equivalent but is not guaranteed to be bitwise identical for every solver and platform.

### Testing

The following coverage was added or performed:

- Checkpoint HDF5 round trips with multiple particle containers
- Both charge/mass storage modes
- Optional spin state
- Cavity-phase metadata
- Tracking-schedule restart positioning
- Diagnostic HDF5 rewind and in-place append
- Checkpoint/restart with phase-space HDF5 disabled
- Same-rank uninterrupted versus restarted execution
- Repeated restart from an older checkpoint
- One-rank checkpoint restarted on two ranks
- Two-rank checkpoint restarted on one rank

Results:

- `TestCheckpointFile` passes.
- `TestStepSizeConfig` passes.
- All 14 `TestMultiContainerPartBunch` tests pass on one rank.
- Same-rank `.stat` and `.h5` output is identical to an uninterrupted run.
- Cross-rank HDF5 output agrees within an absolute tolerance of `1e-12`.
- HDF5-disabled restart reproduces the uninterrupted statistics exactly.
- Repeated restart produces no duplicate diagnostic records.

### Documentation

The OPALX manual now documents:

- The `CHECKPOINTFREQ` option
- The `--restart` command-line argument
- Checkpoint file contents and compatibility checks
- MPI-rank portability
- Diagnostic rewind and append behavior
- Current restart limitations


MichZampetakis:
In a normal scenario, the statistics seem fine. Tested on a A100-interactive node (1 GPU) for the SwissFEL booster without space charge, 2M macroparticles, and 6000 time steps.
image

However, I made two observations:
Another "bad luck" case: I accidentaly stopped my simulation probably just before a `Monitor`. When I tried to restart it, it was stuck, looping over the `Monitor` saving large amounts of data. 

Then, just to clarify: 
In my simulation, I have `MAXSTEPS = 3000, DT = {1.0e-13, 2.0e-12}, ZSTOP={0.20, 13.0};`. I thought that `MAXSTEPS` is indeed the maximum number of timesteps. However, in this case it seems that it is like MAXSTEPS = {3000, 3000} running for 6000 steps in total. Is this intentional? I checked Master and it is the same.

aaadelmann:

aaadelmann:
@MichZampetakis can you try again, and in a little bit the manual should also be updated

aaadelmann:
@rammann please have a look

MichZampetakis:
I tested it and I get the same results as before. 

However I made a few observations:

I did a few runs with two time segments, like Ryan did:
```
TRACK, LINE = SWISSFEL, BEAM = BEAM1, MAXSTEPS = {3000, 3000}, 
			  DT = {1.0e-13, 2.0e-13}, ZSTOP={0.1, 0.35};
```
Regardless of whether I stop in the 1st or 2nd segment, the `--restart` continues correctly. The only minor detail is that if I stop in the second segment, the printed max step is different. 

For example, when the segment was changed, it was printed:
`ParallelTracker::execute[2]> Starting track with dt = 200.000 [fs], track steps = 2999 to 5999.`
Then, I stopped it at timestep 3200, I restarted, and I got: 
`ParallelTracker::execute[2]> Starting track with dt = 200.000 [fs], track steps = 3200 to 6000.`

I checked the `.stat` file but the size is the same for every case. I tried with a checkpoint of 500 steps in case that the `--restart` was saving again the data of the last 400 steps that was not in the checkpoint but I did not observe such thing.


Another issue is with the `Monitor`. I got again into an infinite saving loop:
Screenshot 2026-08-19 at 09 31 34
It seems that it is during the initialization and not during the tracking. I tried to recreate it in a controlled way, but I did not manage. So it might happen due to some kind of corruption when I use `ctrl+C` to kill and `--restart`.

What I did manage to recreate though is that if the normal run stops after the `Monitor` and the `--restart` starts before the `Monitor`, then you get a second shot in the `Monitor.h5`. So, in my case instead of getting a file of ~360 MB, I was getting a file of 600-800 MB. Not sure if this is an issue with the `--restart` functionality that should delete the previous `.h5` or if `Monitor` should have a flag to not save twice in such cases. On the othe hand, it might as well serve as a feature to be able to crosscheck the two shots :)

aaadelmann:
@MichZampetakis  Both problems should be fixed

MichZampetakis:
Now the track steps seem to be consistent, no matter in which segment I `--restart`:
```
ParallelTracker::execute[2]> Starting track with dt = 100.000 [fs], track steps = 0 to 3000.
...
ParallelTracker::execute[2]> Starting track with dt = 200.000 [fs], track steps = 3000 to 6000.
```

However, I am still having issues with the `Monitor` when I `--restart`. It seems that it still saves extra shots, even when the `--restart` run is initialized. Strange enough, `M2` also records data during initialization, even though my simulations do not even reach its position. `M2` is at 0.3 m while the simulation stops at 0.257 m.
```
Ippl> * **********************************************************************************
Ippl> 
OPAL-X> * OrbitThreader maxDistance= 1.79769e+308
OPAL-X> * OrbitThreader #elements  = 2
OPAL-X[2]> Save 'M1.h5'
OPAL-X[2]> Save 'M1.h5'
OPAL-X> * OrbitThreader maxDistance= 1.79769e+308
OPAL-X> * OrbitThreader #elements  = 3
...
OPAL-X[2]> Save 'M2.h5'
OPAL-X[2]> Save 'M2.h5'
OPAL-X> * OrbitThreader maxDistance= 1.79769e+308
OPAL-X> * OrbitThreader #elements  = 2
...

OPAL-X> * Creating new file for data: 'data/swissfel-booster_ElementPositions.sdds'
OPAL-X> * Track start at: 10:22:31, t= 340.000 [ps]; sStart at: 95.626 [mm]
OPAL-X> * Initial dt = 200.000 [fs]
OPAL-X> * Max integration steps = 6000, next step = 3200
OPAL-X>                dt [ns]           sStop [m]        num Steps [1]
OPAL-X>            1.000000e-13        1.000000e-01                3000
OPAL-X> ->         2.000000e-13        3.500000e-01                3000
PartBunch::performBunchSanityChecks[2]> ========= Done performing PartBunch sanity checks... =========
ParallelTracker::execute[2]> Starting track with dt = 200.000 [fs], track steps = 3200 to 6000.
OPAL-X> * 10:22:31 Step   3200 container[0] at 95.686 [mm], t= 340.200 [ps], E=4.722 [MeV]
OPAL-X[2]> Save 'M1.h5'
OPAL-X[2]> Save 'M1.h5'
...
```
If I do a whole run, then `M2.h5`does not appear.

This is the `.in` I am using for reference:
[swissfel-booster.in.rtf](https://github.com/user-attachments/files/31297698/swissfel-booster.in.rtf)

MichZampetakis:
_(Apparently, my comment was never saved. I hope this will not be a dublicate)_

Now the time steps seem to be consistent no matter in which segment I `--restart`:
```
ParallelTracker::execute[2]> Starting track with dt = 100.000 [fs], track steps = 0 to 3000.
...
ParallelTracker::execute[2]> Starting track with dt = 200.000 [fs], track steps = 3000 to 6000.
```

However, I am still having issues with the `Monitor`. I still get some extra shots that seem to be during initialization when I `--restart`. I am not sure then if I get another shot while running, or if the old one is overwriten, But I get the `Save 'M1.h5'` message.

Strange enough, even the second `Monitor` records data during initialization, which is much further downstream. Important to note is that this simulations does not even reach `M2` which is located at 0.3 m, while the simulation stops at 0.257 m.
```
OPAL-X> * OrbitThreader maxDistance= 1.79769e+308
OPAL-X> * OrbitThreader #elements  = 2
OPAL-X[2]> Save 'M1.h5'
OPAL-X[2]> Save 'M1.h5'
OPAL-X> * OrbitThreader maxDistance= 1.79769e+308
OPAL-X> * OrbitThreader #elements  = 3
...
OPAL-X[2]> Save 'M2.h5'
OPAL-X[2]> Save 'M2.h5'
OPAL-X> * OrbitThreader maxDistance= 1.79769e+308
OPAL-X> * OrbitThreader #elements  = 2
...
OPAL-X> * Max integration steps = 6000, next step = 3200
OPAL-X>                dt [ns]           sStop [m]        num Steps [1]
OPAL-X>            1.000000e-13        1.000000e-01                3000
OPAL-X> ->         2.000000e-13        3.500000e-01                3000
PartBunch::performBunchSanityChecks[2]> ========= Done performing PartBunch sanity checks... =========
ParallelTracker::execute[2]> Starting track with dt = 200.000 [fs], track steps = 3200 to 6000.
OPAL-X> * 10:55:37 Step   3200 container[0] at 95.686 [mm], t= 340.200 [ps], E=4.722 [MeV]
OPAL-X[2]> Save 'M1.h5'
OPAL-X[2]> Save 'M1.h5'
ParallelTracker::execute[2]> Wrote checkpoint 'swissfel-booster_checkpoint.h5' after global step 3300.
...
```
If I run a single run without `--restart`, then `M1.h5` seems fine and `M2.h5` does not exist!

For reference, this is my `.in` file:
[swissfel-booster.in.rtf](https://github.com/user-attachments/files/31298736/swissfel-booster.in.rtf)

MichZampetakis:
As we discussed, the `Monitor` data dump during initialization is now fixed.

However, I still get some dublicate recordings if the first run stops after the `Monitor` but the `--restart` starts before the `Monitor`.

```
OPAL-X> * Max integration steps = 3199, next step = 3100
OPAL-X>                dt [ns]           sStop [m]        num Steps [1]
OPAL-X>            1.000000e-13        1.000000e-01                3000
OPAL-X> ->         2.000000e-13        3.500000e-01                 199
PartBunch::performBunchSanityChecks[2]> ========= Done performing PartBunch sanity checks... =========
ParallelTracker::execute[2]> Starting track with dt = 200.000 [fs], track steps = 3100 to 3199.
OPAL-X> * 14:20:57 Step   3100 container[0] at 95.745 [mm], t= 320.200 [ps], E=6.600 [MeV]
OPAL-X[2]> Save 'M1.h5'
OPAL-X[2]> Save 'M1.h5'
OPAL-X[2]> * Dump phase space of last step
```

This is the input file that I am using (no fieldmaps required):
[monitor_test.txt](https://github.com/user-attachments/files/31374885/swissfel-booster.txt)

It is set up so that it records the last checkpoint at time step 3100, then `M1.h5` is recorded and stops at time step 3199.
Then, it restarts at time step 3100 and a second shot is probably appended in `M1.h5`.

MichZampetakis:
With the latest commits, the `--restart` run seems to save again the `M1.h5` file, but the previous one is erased.

However, now I noticed that the `M1.h5` is half the size that was in the past. It was ~320 MB, now I get ~160 MB. This might be because before the monitor was recording `Step#0` and `Step#1`, even for a single passage. I think this was related to the reference particle check that it was doing and it was saving an extra step. Now this seems to be fixed, and only `Step#0` appears in the `M1.h5`.

Probably we also need to be careful in the future, when we will use a `Monitor` in a synchrotron, and multiple passages will take place.

src/Main.cpp
```diff
@@ -329,10 +363,12 @@ int main(int argc, char* argv[]) {
 
             if (opal->inRestartRun()) {
                 if (restartFileName.empty()) {
-                    restartFileName = opal->getInputBasename() + std::string(".h5");
+                    fs::path restartPath(fname);
+                    restartPath.replace_extension("");
+                    restartFileName = restartPath.string() + "_restart.h5";
```
MichZampetakis: Shouldn't it be `"_checkpoint.h5"` to follow the checkpoint filename?

src/Algorithms/ParallelTracker.cpp
```diff
@@ -62,6 +63,37 @@
...
+                active_m = false;
+            }
+        }
+
+    private:
+        OpalData& opalData_m;
+        bool active_m{true};
+    };
+}  // namespace
+
```
rammann: I think from a code structure point of view, it would make sense to have this functionality baked into the `OpalData` class, instead of wrapping it. 

Inside `OpalData` we already have the functions
```
    /// true if in follow-up track
    bool hasPriorTrack();

    /// true if in follow-up track
    void setPriorTrack(const bool& value = true);

    /// true if we do a restart run
    bool inRestartRun();

    /// set OPAL in restart mode
    void setRestartRun(const bool& value = true);

    /// store the location where to restart
    void setRestartStep(int s);
```

aaadelmann: good point will look into it!

aaadelmann: Is easy possible -- good catch!

aaadelmann: implemented

aaadelmann:

rammann (changes requested):
I think we should remove the old (and now unused) restart mechanism in the `OpalData` class to avoid potential future confusion.

This includes functions like `OpalData::getRestartStep()` and `OpalData::setRestartStep()`.

rammann (approved):

MichZampetakis (approved):

aaadelmann:

aaadelmann:

aaadelmann:

aaadelmann:

PR #526, opened 2026-08-17
Implement Collimator Element

### LINKED ISSUE #488: Collimator Element Missing
OPALX is missing a working collimator element. This is the last element (apart from the bends #485) that is required to simulate MuE4 in the configuration without the Wien filter. 

Implementation of a collimator is simple, no fields are applied and particles outside the aperture need to be deleted. The logic used in `ParticleContainer::markParticlesOutside(double)` can be reused.

rammann:
## Summary
This PR introduces a basic collimator element. 
For now, it is a drift element with a specified aperture. In future work this will be extended to allow for more complex apertures and crucially, local physics processes. 
Closes #488.

## Testing
Unit and regression tests are passing on CPU and GPU.


src/AbsBeamline/Collimator.cpp
```diff
@@ -0,0 +1,43 @@
+//
+// Class Collimator
+//   Interface for a transverse-aperture collimator.
```
aaadelmann, line 3: why only transverse?

aaadelmann (changes requested):
A collimator should not be restricted to one dimension i.e. transverse

src/AbsBeamline/Collimator.cpp
rammann: Yes agreed. This is a starting point for implementing a more complex collimator. For now all its functionality is inherited from `ElementBase`, so it is simply a drift with an aperture. In a next step we can define more complex behavior (like longitudinally changing aperture) by changing its own `apply()` function and the allowed apertures. 

But I think it makes sense to start from a simple element (which works) and iteratively add functionality.

aaadelmann (approved):

PR #528, opened 2026-08-18
Recreate broken NightlyBuildX tools environment when Python is unimportable

biddisco:
This should fix regression testing pipeline on cscs/alps


aaadelmann (approved):

biddisco:
cscs-ci run cscs-ci-openmp

biddisco:
Regression tests seem to be working, but before/when merging this branch, be aware that we are now using the NightlyBuildX branch here https://github.com/OPALX-project/NightlyBuildX/pull/8
Note that if the BuildX branch is merged and eventually updated, we will need to update the branch used in this PR accordingly.

biddisco:
Since the NightlyBuildX repo has had the noglp and alps-ci branches merged into master, I've switched this PR to use master branch from there. I'll let the CI cycle and if all passes then this is good to merge.

PR #529, opened 2026-08-18
Add G4BL 2D Magnetostatic Fieldmap to Solenoid

rammann:
## Summary
This PR adds the capability for OPALX to directly read the G4BL 2DMagnetostatic fieldmap format.
For full support of this format a new variable `zRevers_m`, which inverts the fieldmap values in z direction while leaving the radial component unchanged, had to be added to the `Fieldmap` class. For now this is only enabled for the `G4BL2DMagnetostatic` format.

### Calculating Transfer Matrices
The first test was calculating and comparing the transfer matrices of the same solenoid fieldmap we get in OPALX and G4BL. It is the fieldmap of the solenoid at the beginning of the MuE4 secondary beamline. 
The testing setup was a monitor before and after the solenoid and 13 test particles which were used to calculate the transfer matrices.
image
The transfer matrices and their difference:
image
We can also see agreement between the on-axis field values
image
as well as the test particles position at the second monitor:
image

### Phase Space Comparison
We can also visually compare the transversal phase space for a larger run with 15000 particles.
In this case we have 8 recording planes (monitors) throughout the lattice. 
image

#### z=150mm
image

#### z=600mm
image

#### z=1050mm
image

#### z=1500mm
image

#### z=1950mm
image

#### z=2400mm
image

#### z=2850mm
image

#### z=3450mm
image

## Regression and Unit Tests
Todo


aaadelmann:
Nice documentation of the work! A few histograms showing both G4Bl and OPALX would be nice.

src/AbsBeamline/Solenoid.h
```diff
@@ -122,6 +122,17 @@ class Solenoid : public ElementBase {
...
+    /**
+     * @brief Read the field map back to front.
+     *
+     * Mirrors the map in z and negates Bz, leaving Br alone -- what turning the
+     * magnet around does to an axisymmetric field. Only G4beamline cylinder
+     * maps support this; other map types throw when it is set.
+     */
+    void setZReverse(bool zReverse);
+
+    bool getZReverse() const;
```
aaadelmann: would isZReverese describe the function better?

aaadelmann (approved):

src/AbsBeamline/Solenoid.h
rammann: Yes sounds good, I'll change it before I merge.

rammann:
> Nice documentation of the work! A few histograms showing both G4Bl and OPALX would be nice.

Sure, I can add some histograms too.

rammann:
Here the histograms of the transversal phase space coordinates at the first and last monitors. OPALX in blue and G4BL in red (not visible because of perfect overlap).
image

aaadelmann (approved):

rammann:
Regression tests and unit tests passing on Daint. Merging once the CI runs through.

